PR #116, opened 2020-05-22
[HIPIFY] Set configurable install path

paulfreddy:
Add capability to build package as per user choice.
If the parameter CPACK_PACKAGING_INSTALL_PREFIXX is not provided, 
default to /opt/rocm/hip

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,11 @@ if (MSVC AND MSVC_VERSION VERSION_LESS "1900")
     return()
 endif()
 
+#Set configurable install path
+if (UNIX)
+    set(CPACK_PACKAGING_INSTALL_PREFIX "/opt/rocm/hip" CACHE PATH "Package Installation path for HIP")
+endif (UNIX)
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ if (MSVC AND MSVC_VERSION VERSION_LESS "1900")
     return()
 endif()
 
+#Set configurable install path
+set(CPACK_PACKAGING_INSTALL_PREFIX "/opt/rocm/hip" CACHE PATH "Package Installation path for HIP")
```
emankov, line 11: Packaging is implemented only for Linuxes, whereas hipify-clang could be built for Windows as well. Thus, move this set down inside `if (UNIX)` condition, please.

emankov: 1. Currently, right. 2. In order not to change your line(s) when we start to support packaging for Windows. And why should the user see the Linux paths in its CMakeCache file on Windows? )

paulfreddy: Done

PR #141, opened 2020-06-23
Change the hipify packaging target

aakanksha555:
Simplify packaging target. The command used by end user will be "$make package" now.
Also, simplify code to name the DEB and RPM package to *-amd64.deb and *.x86_64.rpm respectively.


CMakeLists.txt
```diff
@@ -118,25 +118,25 @@ install(
   PATTERN "openmp_wrappers" EXCLUDE)
 
 if (UNIX)
+    if(POLICY CMP0037)
+        cmake_policy(PUSH)
+        cmake_policy(SET CMP0037 OLD)
+    endif()
```
emankov, line 124: Why CMP0037 cmake policy is introduced once again?

aakanksha555: CMake 2.8.12 and lower allowed unrestricted choice for the target name. CMP0037 OLD cmake policy allows the use of restricted target names such as 'target'.

emankov: We've already discussed it: hipify-clang has a restriction on cmake version (see above):
```cmake
cmake_minimum_required(VERSION 3.5.1)
```

aakanksha555: We don't have to use the lower cmake version. The policy allows the use of restcricted target names.

emankov: So, you need `CMP0037 OLD` policy only to use `package` as a target name instead of `pkg_hipify-clang`, right?
Maybe better not to use a reserved word `package`?

emankov: @searlmc1, Do we need exactly `package` as a target name?

emankov: @searlmc1, @aakanksha555, any updates?

aakanksha555:
Working on a new patch to use the target "package" without the use of the policy CMP0037

PR #151, opened 2020-07-09
[HIPIFY][package] Set packaging arch, field separators

nelsonc-amd:
Removed work-around and use CPACK_PROJECT_CONFIG_FILE instead.
This enables the correct application of underscores for debian.


nelsonc-amd:
@pruthvistony

Issue #178, opened 2020-08-26
[HIPIFY][CUDA] Starting from CUDA 11.0 Update 1 "triple less-than sign" leads to an error


emankov:
The error reported is **`Use of undeclared identifier cudaConfigureCall`**:

```cuda
axpy_kernel <<<1, 2>>> (a, y, x);
            ^
```
There is no such error when compiling by clang against CUDA 11.0 (before Update 1) and previous versions.

The error occurs when compiling CUDA code containing kernel launch in `CUDAKernelCallExpr` syntax by all versions of clang, which initially supported CUDA 11.0: 10.x - 11.x, 12.0.0git (trunk).
**Looks like an issue with clang.**

The following tests are failed:

_unit_tests/compilation_database/cd_intro.cu
unit_tests/device/atomics.cu
unit_tests/device/device_symbols.cu
unit_tests/device/math_functions.cu
unit_tests/kernel_launch/kernel_launch_01.cu
unit_tests/libraries/CUB/cub_01.cu
unit_tests/libraries/CUB/cub_02.cu
unit_tests/libraries/CUB/cub_03.cu
unit_tests/libraries/cuRAND/benchmark_curand_kernel.cpp
unit_tests/libraries/cuRAND/poisson_api_example.cu
unit_tests/namespace/ns_kernel_launch.cu
unit_tests/pp/pp_if_else_conditionals.cu
unit_tests/pp/pp_if_else_conditionals_01.cu
unit_tests/pp/pp_if_else_conditionals_01_LLVM_10.cu
unit_tests/pp/pp_if_else_conditionals_LLVM_10.cu
unit_tests/samples/2_Cookbook/0_MatrixTranspose/MatrixTranspose.cpp
unit_tests/samples/2_Cookbook/13_occupancy/occupancy.cpp
unit_tests/samples/2_Cookbook/1_hipEvent/hipEvent.cpp
unit_tests/samples/2_Cookbook/2_Profiler/Profiler.cpp
unit_tests/samples/2_Cookbook/7_streams/stream.cpp
unit_tests/samples/2_Cookbook/8_peer2peer/peer2peer.cpp
unit_tests/samples/MallocManaged.cpp
unit_tests/samples/allocators.cu
unit_tests/samples/axpy.cu
unit_tests/samples/coalescing.cu
unit_tests/samples/cudaRegister.cu
unit_tests/samples/dynamic_shared_memory.cu
unit_tests/samples/intro.cu
[...]

PR #206, opened 2020-10-20
[HIPIFY][#178][patch] Add patches for clang 10.0.0, 10.0.1, and 11.0.0

emankov:
+ Download the patch and unpack it into your LLVM source directory: the file Cuda.cpp will be overwritten; needs further rebuilding of LLVM
+ Update README.md accordingly

The fix will arrive with LLVM 11.0.1 and 12.0.0.


README.md
```diff
@@ -47,23 +47,102 @@ After applying all the matchers, the output HIP source is produced.
...
-| [9.0.0](http://releases.llvm.org/download.html#9.0.0)   , [9.0.1](http://releases.llvm.org/download.html#9.0.1)    | [10.1](https://developer.nvidia.com/cuda-10.1-download-archive-base)     | +           | +         |
-| [10.0.0](http://releases.llvm.org/download.html#10.0.0) , [10.0.1](https://releases.llvm.org/download.html#10.0.1) | [**11.0**](https://developer.nvidia.com/cuda-11.0-download-archive), [11.0 Update 1](https://developer.nvidia.com/cuda-11.0-update1-download-archive)\*\*\*, [11.1](https://developer.nvidia.com/cuda-downloads)\*\*\* | +           | +         |
-| [**11.0.0**](https://releases.llvm.org/download.html#11.0.0) | [**11.0**](https://developer.nvidia.com/cuda-11.0-download-archive), [11.0 Update 1](https://developer.nvidia.com/cuda-11.0-update1-download-archive)\*\*\*, [11.1](https://developer.nvidia.com/cuda-downloads)\*\*\* | + <br/> **LATEST STABLE RELEASE** | + <br/> **LATEST STABLE RELEASE** |
+<table align="center">
```
yxsamliu, line 50: It used to be md format but now becomes html. Is this intended?

emankov: It is intended, cause I was unable to apply HTML's `colspan` and `rowspan` analogues in GitHub's Markdown format (`||` doesn't work). Nonetheless, GitHub correctly parses Markdown documents with simple HTML injections.
Before (mostly in md format):
before
After (with HTML table injection)
after

PR #229, opened 2020-11-19
[HIPIFY] Standardizing Package name

ashutom:
Enables standards compliant package naming for debian and rpm

Signed-off-by: Ashutosh Mishra <ashutosh.mishra@amd.com>


CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ if (MSVC AND MSVC_VERSION VERSION_LESS "1900")
     return()
 endif()
 
-find_package(LLVM REQUIRED)
+find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH} /opt/rocm )
```
emankov, line 10: OS-specific code should go separately.

packaging/hipify-clang.txt
```diff
@@ -7,25 +7,53 @@ install(DIRECTORY @HIPIFY_INSTALL_PATH@/include DESTINATION bin)
...
+set(CPACK_INSTALL_PREFIX @CPACK_PACKAGING_INSTALL_PREFIX@)
 set(CPACK_GENERATOR "TGZ;DEB;RPM")
-set(CPACK_BINARY_DEB "ON")
-set(CPACK_BINARY_RPM "ON")
-set(CPACK_RPM_PACKAGE_ARCHITECTURE "${CMAKE_SYSTEM_PROCESSOR}")
+
+#Make proper version for appending
+#Default Value is 99999, setting it first
+set(ROCM_VERSION_FOR_PACKAGE "99999")
+if(DEFINED ENV{ROCM_LIBPATCH_VERSION})
```
emankov, line 25: Which component provides this env. variable?

emankov (changes requested):

packaging/hipify-clang.txt
ashutom: This is defined by the CI builds env. When User builds locally these are undefined & hence high number for version is replaced

CMakeLists.txt
ashutom: @emankov : you mean /opt/rocm? Let me update

ashutom: @emankov : Updated new commit

emankov: I meant `if (UNIX)`, actually

ashutom: @emankov, the current code : find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH}) , is  OS agnostic now.
This shall run perfectly fine on any OS. Its is just trying to search as given by the CMAKE_PREFIX_PATH while compiling. It also shall search for the usual locations as well.
The last code was specific to rocm and was relying that the rocm is installed on the system (in linux /opt/rocm) hence it would get the path from there.
Request you to kindly review.

emankov: Yes, but initially you wanted a default value for cases when `CMAKE_PREFIX_PATH` is not provided, right? Why did you decide to remove it?

ashutom: Yes, because you reviewed that code should be OS agnostic.  latest commit new code is adapting both now

emankov: I see `if (NOT ...)` you like more. Well, do not mind, but `if (...)` is shorter, nonetheless.

emankov (approved):

PR #315, opened 2021-04-12
[HIPIFY] Address sanitizer support

aakanksha555:
SWDEV-269088 - [ROCm Feature Address Sanitizer Support] address sanitizer needs to be enabled in hipify-clang build


emankov (changes requested):
Why you need the AddressSanitizer unconditionally?
What if ADDRESS_SANITIZER is false?
Have you run the tests? They all are failed (at least on Windows).

aakanksha555:
>Why you need the AddressSanitizer unconditionally?
> What if ADDRESS_SANITIZER is false?
Got it, I'll add an else bit.

> Have you run the tests? They all are failed (at least on Windows).
Would it be possible for you to share an error log you see on Windows?

emankov:
> Would it be possible for you to share an error log you see on Windows?

It is because of missing `else` and as a result - missing `-std=c++`

emankov:
Where does `ADDRESS_SANITIZER` define come from?
And I'd prefer that it will be switched off in Release.

aakanksha555:
> Where does `ADDRESS_SANITIZER` define come from?
> And I'd prefer that it will be switched off in Release.

The ADDRESS_SANITIZER option may be specified as a cmake option to enable/ disable address sanitizer; it will be disabled by default unless explicitly enabled using the cmake option "-DADDRESS_SANITIZER=ON".
With ADDRESS_SANITIZER enabled all tests are failing with "ERROR: AddressSanitizer: use-after-poison .." error, potentially exposing memory leaks. We currently do not have the tests enabled in our internal CI so it won't cause an issue even if this patch is pushed through; but this should definitely be looked into.

CMakeLists.txt
```diff
@@ -85,8 +85,15 @@ else()
...
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${EXTRA_CFLAGS}")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${EXTRA_CFLAGS} ${StdCpp} -DHIPIFY_CLANG_RES=\\\"${LLVM_LIBRARY_DIRS}/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}\\\"")
+# Address Sanitize Flag
+if (ADDRESS_SANITIZER)
+    set(addr_var address)
+else()
+    set(addr_var )
+endif()
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${EXTRA_CFLAGS} -fsanitize=${addr_var}")
```
emankov, line 95: `-fsanitize=` is useless if not erroneous in case of not `ADDRESS_SANITIZER`.

emankov (changes requested):

emankov (changes requested):
Building after cmake with `-DADDRESS_SANITIZER=1` leads to the following error on Windows:

```
dist\include\llvm/Support/Compiler.h(434,11): fatal error C1083: Cannot open include file: 'sanitizer/asan_interface.h': No such file or directory
```
Additional include path to LLVM sanitizer should be specified for building hipify-clang.
In LLVM sources `asan_interface.h` lives here `llvm-project\compiler-rt\include\sanitizer\`.

Please, also check if compiler-rt should be built as well. Currently, only clang is needed for building hipify-clang besides llvm. If some `compiler-rt` libs are needed during building hipify-clang with `-fsanitize=address`, then the include path should be specified to `asan_interface.h` from `dist` folder, not `llvm-project`.

Finally, README should be updated accordingly.

aakanksha555:
> Building after cmake with `-DADDRESS_SANITIZER=1` leads to the following error on Windows:
> 
> ```
> dist\include\llvm/Support/Compiler.h(434,11): fatal error C1083: Cannot open include file: 'sanitizer/asan_interface.h': No such file or directory
> ```
> 
> Additional include path to LLVM sanitizer should be specified for building hipify-clang.
> In LLVM sources `asan_interface.h` lives here `llvm-project\compiler-rt\include\sanitizer\`.
> 
> Please, also check if compiler-rt should be built as well. Currently, only clang is needed for building hipify-clang besides llvm. If some `compiler-rt` libs are needed during building hipify-clang with `-fsanitize=address`, then the include path should be specified to `asan_interface.h` from `dist` folder, not `llvm-project`.
> 
> Finally, README should be updated accordingly.

I think I always had my llvm built with compiler-rt so I hadn't seen this error before. But, I verified the error by building HIPIFY with llvm built with and without compiler-rt. Including compiler-rt when building llvm does the trick.

emankov:
What about this (at least on Windows):

> Additional include path to LLVM sanitizer should be specified for building hipify-clang.

And what about libs, do we need to specify the lib(s) as well while building `hipify-clang`?

aakanksha555:
I don't have a windows setup. Would it be possible for you to provide some assistance with that?
Currently, without any additional paths, what do you see if you try to build on Windows?

emankov:
> I don't have a windows setup. Would it be possible for you to provide some assistance with that?
> Currently, without any additional paths, what do you see if you try to build on Windows?

Please, see the compilation error above.

emankov:
> > > I don't have a windows setup. Would it be possible for you to provide some assistance with that?
> > > Currently, without any additional paths, what do you see if you try to build on Windows?
> > 
> > 
> > Please, see the compilation error above.
> 
> Even after you build/install compiler-rt with llvm? On linux, hipify picks it up from the dist folder at "llvm-project/dist/lib/clang/13.0.0/include/sanitizer/asan_interface.h"

I need time to check. Anyway, by making changes in the build system, you should be able to check the build on all supported platforms. There is a detailed instruction in [Readme.md](https://github.com/ROCm-Developer-Tools/HIPIFY#readme) on how to build and test on Windows.

emankov (approved):
`asan_interface.h` is found if `compiler-rt` is specified when building `LLVM` on Windows. hipify-clang build with `-DADDRESS_SANITIZER=1` on Windows is correct now, all tests passed.

So, the change LGTM.

PR #390, opened 2021-08-27
[HIPIFY] Add dependency on rocm-core

icarus-sparry:
The intention is to make all the rocm packages depend on rocm-core to
enable eveerything to be removed cleanly by just removing the
rocm-core package.

Signed-off-by: Icarus Sparry <icarus.sparry@amd.com>

### DIFF
--- a/packaging/hipify-clang.txt
+++ b/packaging/hipify-clang.txt
@@ -4,6 +4,10 @@ project(hipify-clang)
 install(PROGRAMS @HIPIFY_INSTALL_PATH@/hipify-clang DESTINATION bin)
 install(DIRECTORY @HIPIFY_INSTALL_PATH@/include DESTINATION bin)
 
+# hipify-clang.txt is processed to produce CMakeList.txt,
+# pass down value from parent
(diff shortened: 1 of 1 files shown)


emankov:
HIPIFY package is a standalone tool independent from rocm-core.

icarus-sparry:
The point of "rocm-core" is to provide an easy way to remove all of rocm. So the question is "Is hipify part of rocm?" If installing rocm is going to install hipify then removing rocm should remove hipify. I may not agree with it but the method that has been decided upon is to make all packages which make up rocm to depend on essentially a dummy package called "rocm-core" so removing rocm-core causes the package managers to remove all of rocm.

The alternative is to say hipify is not part of rocm.

emankov:
> So the question is "Is hipify part of rocm?" 

Only theoretically. It is not installed with ROCm HIP; it has its own independent package. 
> If installing rocm is going to install hipify then removing rocm should remove hipify.

No, ROCm doesn't install HIPIFY, so it shouldn't remove it.

emankov (commented):
This change is not needed.

emankov (approved):
LGTM, as we agreed on such a dependency for every component of HIP-SDK.

PR #398, opened 2021-09-13
Remove direct dependency on "/opt/rocm" path

dkbhaskaran:
Requesting this change based on internal tickets SWDEV-287841 and SWDEV-287817. These aim to alleviate the issues resulting in assuming the default ROCm path.


emankov (approved):
LGTM

PR #486, opened 2022-01-18
[HIPIFY][SWDEV-316762][fix] Soft link to /opt/rocm/bin is missed for hipify-perl

aakanksha555:
Added a soft link for hipify-perl from /opt/rocm/bin to /opt/rocm/hip/bin


packaging/hipify-clang.rpm_postun
```diff
@@ -0,0 +1,8 @@
+ROCMDIR="/opt/rocm"
+HIPDIR=$ROCMDIR/hip
+HIPBINDIR=$HIPDIR/bin
```
emankov, line 3: HIPBINDIR=/opt/rocm/bin

emankov (commented):
@aakanksha555, have you tested it on those OSs?

aakanksha555:
> @aakanksha555, have you tested it on those OSs?

Tested the .deb packages.

aakanksha555:
Do we want to package all the other scripts in bin as well? 

These a part of rocm 4.5 -
lrwxrwxrwx 1 root root        36 Dec  4 14:14 hipconvertinplace-perl.sh -> ../hip/bin/hipconvertinplace-perl.sh
lrwxrwxrwx 1 root root        31 Dec  4 14:14 hipconvertinplace.sh -> ../hip/bin/hipconvertinplace.sh
]lrwxrwxrwx 1 root root        29 Dec  4 14:14 hipexamine-perl.sh -> ../hip/bin/hipexamine-perl.sh
lrwxrwxrwx 1 root root        24 Dec  4 14:14 hipexamine.sh -> ../hip/bin/hipexamine.sh
lrwxrwxrwx 1 root root        22 Dec  4 14:14 hipify-perl -> ../hip/bin/hipify-perl

emankov:
> Do we want to package all the other scripts in bin as well?
> 
The reported issue was about the soft link to hipify-perl, so the `-perl.sh` scripts might also be needed. As for the scripts without `-perl`, they are hipify-clang based. If you are going to provide the soft links to them, you should provide the soft link to `hipify-clang` too.

emankov:
> > @aakanksha555, have you tested it on those OSs?
> 
> Tested the .deb packages.

I believe the fix has to be tested against all the target OSs.

packaging/hipify-clang.rpm_postun
aakanksha555: Sure, can do that. Was just keeping up with convention in https://github.com/ROCm-Developer-Tools/hipamd/blob/develop/packaging/hip-devel.postinst

aakanksha555:
> > Do we want to package all the other scripts in bin as well?
> 
> The reported issue was about the soft link to hipify-perl, so the `-perl.sh` scripts might also be needed. As for the scripts without `-perl`, they are hipify-clang based. If you are going to provide the soft links to them, you should provide the soft link to `hipify-clang` too.

Sure, I think we should do that. I believe we need coverage for all the components of rocm4.5 package in 5.0 as well.

PR #487, opened 2022-01-19
[HIPIFY][SWDEV-316762][fix] Soft link to /opt/rocm/bin is missed for hipify-perl

aakanksha555:
Revised PR to include sym-links for all hipify-perl and related scripts as well as hipify-clang


packaging/hipify-clang.postinst
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -e
+
+# Soft-link to bin files
+HIPBINDIR="/opt/rocm/hip/bin"
+ROCMBINDIR="/opt/rocm/bin"
```
emankov, line 7: What's the problem in using `CPACK_PACKAGING_INSTALL_PREFIX` set in CMakeLists.txt instead?

CMakeLists.txt
```diff
@@ -115,6 +115,10 @@ if(UNIX)
     set(CPACK_PACKAGING_INSTALL_PREFIX "/opt/rocm/hip" CACHE PATH "HIP Package Installation Path")
```
emankov, line 115: I'd add the following two lines here:
```cmake
set(HIPBINDIR ${CPACK_PACKAGING_INSTALL_PREFIX}/bin)
set(ROCMBINDIR ${CPACK_PACKAGING_INSTALL_PREFIX}/../bin)
```

And remove the following two lines in the below `packaging/hipify-clang.*` files:
```cmake
HIPBINDIR="/opt/rocm/hip/bin"
ROCMBINDIR="/opt/rocm/bin"
```

Could you test it on all the target OSs, please?
Thank you!

aakanksha555: Tested it out. The post install files don't have visibility of the local variables in CMakeLists.txt.

emankov: Aha. I thought that not only `hipify-clang.txt` does have visibility of local variables declared in `CMakeLists.txt`.

PR #514, opened 2022-03-12
[HIPIFY][package] Delete hip folder after uninstallation if empty

aakanksha555:
Delete hip folder after uninstall; ignore if not empty.
Temporary workaround so that hip folder isn't left in /opt/rocm-x.y.z path after driver uninstall

### DIFF
--- a/packaging/hipify-clang.rpm_postun
+++ b/packaging/hipify-clang.rpm_postun
@@ -1,5 +1,3 @@
-HIPDIR=@CPACK_PACKAGING_INSTALL_PREFIX@
-
 if [ $1 -eq 0 ]; then
   rm -f @ROCMBINDIR@/hipify-perl
   rm -f @ROCMBINDIR@/hipify-clang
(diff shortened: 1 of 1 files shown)


packaging/hipify-clang.rpm_postun
```diff
@@ -11,4 +10,6 @@ if [ $1 -eq 0 ]; then
   rm -f @ROCMBINDIR@/findcode.sh
   rm -f @ROCMBINDIR@/finduncodep.sh
   rmdir --ignore-fail-on-non-empty @ROCMBINDIR@
+  rmdir --ignore-fail-on-non-empty @HIPBINDIR@
+  rmdir --ignore-fail-on-non-empty ${HIPDIR}
```
emankov, line 14: You may use `@CPACK_PACKAGING_INSTALL_PREFIX@` here instead of `HIPDIR` as it is obviously the HIP/HIPIFY root installation dir.

emankov (approved):
Thanks!

PR #528, opened 2022-04-12
[HIPIFY][build]File reorganization changes

raramakr:
Header files installed in /opt/rocm-ver/include/hipify
Public bash scripts installed in /opt/rocm-ver/bin
Private bash scripts-findcode.sh finduncodep.sh  installed in /opt/rocm-ver/libexec/hipify
Softlink for backward compatibility


packaging/hipify-clang.txt
```diff
@@ -2,10 +2,10 @@ cmake_minimum_required(VERSION 3.6.3)
 project(hipify-clang)
 
 install(PROGRAMS @HIPIFY_INSTALL_PATH@/hipify-clang DESTINATION bin)
-install(DIRECTORY @HIPIFY_INSTALL_PATH@/include DESTINATION bin)
+install(DIRECTORY @HIPIFY_INSTALL_PATH@/include/ DESTINATION include/hipify)
```
emankov, line 5: I don't see softlink for backward compatibility for the `include` subfolder with clang's system header files. Current implementation of `hipify-clang` binary awaits those header files in `bin/include`.
FMPOV, we need two additional things here: 1. To change `hipify-clang` to use header files from `include/hipify` (in progress, will be done in a separate change) 2. To add a corresponding soflink to the former `bin/include` location.

Otherwise, hipification process by `hipify-clang` will be broken.

PR #532, opened 2022-04-22
[HIPIFY][#531][SWDEV_331863][temporary][workaround] Add a workaround for the `SWDEV_331863` blocker

emankov:
+ To build hipify-clang against `trunk LLVM (15.0.0git)` nothing to do else
+ To build hipify-clang against `trunk LLVM (15.0.0git)` older than `d79ad2f1dbc2db63121620f55d6cfa915f2733ac` (2022.04.11, DR: https://reviews.llvm.org/D123574) `-DSWDEV_331863=ON` should be added to hipify-clang's cmake command line

[ToDo] Remove `SWDEV_331863` related guards from `CMakeLists.txt` and `HipifyAction.cpp` when the blocker `SWDEV_331863` is overcome

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,12 @@ else()
     set(addr_var )
 endif()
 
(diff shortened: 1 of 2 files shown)


searlmc1 (approved):
LGTM

PR #588, opened 2022-07-06
[HIPIFY][build] Remove workaround code to strip destination directory hip

raramakr:
Hipify build script updated, so that it doesn't use "hip" as destination in cpack packaging prefix
The workaround in source code is not at all required


nunnikri (approved):
Need to remove this work around asp. This is changing the out directory which is a very risky one

emankov (commented):
LGTM!

PR #678, opened 2022-10-28
[HIPIFY][build][SWDEV-364233][fix] Wrong `RUNPATH` in hipify-clang executable

ashutom:
Correcting runpath of hipify-clang from
runpath to correct rpath

Signed-off-by: Ashutosh Mishra <ashutosh.mishra@amd.com>

Changes : 

1) Updated the link directories by removing 
link_directories(${LLVM_LIBRARY_DIRS})
and adding target_link_directories(hipify-clang PRIVATE ${LLVM_LIBRARY_DIRS})  because the documentation ( https://cmake.org/cmake/help/latest/command/link_directories.html) states that : " This command is rarely necessary and should be avoided where there are other choices"

2) Corrected the RPATH
     by removing the INSTALL_RPATH and using --disable-new-dtags

Tested for the corrected rpath

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,12 +24,12 @@ if(MSVC AND MSVC_VERSION VERSION_LESS "1900")
 endif()
 
 include_directories(${LLVM_INCLUDE_DIRS})
-link_directories(${LLVM_LIBRARY_DIRS})
 add_definitions(${LLVM_DEFINITIONS})
 
 file(GLOB_RECURSE HIPIFY_SOURCES src/*.cpp)
 file(GLOB_RECURSE HIPIFY_HEADERS src/*.h)
 add_llvm_executable(hipify-clang ${HIPIFY_SOURCES} ${HIPIFY_HEADERS})
+target_link_directories(hipify-clang PRIVATE ${LLVM_LIBRARY_DIRS})
 
 set(CMAKE_CXX_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang++)
 set(CMAKE_C_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang)
@@ -131,7 +131,14 @@ install(
     PATTERN "openmp_wrappers" EXCLUDE)
 
 option(FILE_REORG_BACKWARD_COMPATIBILITY "Enable File Reorg with backward compatibility" ON)
+
 if(UNIX)
+
+    #get rid of any RPATH definations already
+    set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "")
+    #set RPATH for the binary
+    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--disable-new-dtags -Wl,--rpath,$ORIGIN/../lib" )
+
     if(FILE_REORG_BACKWARD_COMPATIBILITY)
         include(hipify-backward-compat.cmake)
     endif()


solaiys:
LGTM

emankov:
Hello @ashutom,

Could you please shed more light on what exactly was wrong? What issue was fixed with your change?
And as for the testing, on what operating systems were you testing the change? Did you run unit tests also?

ashutom:
> 

@emankov 

Currently ( without this patch ) the binary produced has the following information :
```
readelf -d amexp/opt/rocm-5.5.0-9999/bin/hipify-clang

Dynamic section at offset 0x2f15b38 contains 34 entries:
  Tag        Type                         Name/Value
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/../lib:/home/taccuser/compute/out/ubuntu-20.04/20.04/llvm/lib]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [librt.so.1]
```

As you can see the binary has RUNPATH and not RPATH,  also the path itself contains the information about the build system where it was compiled. This is unnecessary leak of of data and against our policies.

After the patch :   
```
readelf -d amexp/opt/rocm-5.5.0-9999/bin/hipify-clang

Dynamic section at offset 0x2f15af8 contains 34 entries:
  Tag        Type                         Name/Value
 0x000000000000000f (RPATH)              Library rpath: [/opt/rocm-5.5.0-9999/lib]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]


``` 

This has RPATH instead of RUNPATH which has better enforcement also the patch is clean one without any unnecessary data

This change is effective for all the OS distributions

emankov:
Thanks, @ashutom for both your explanations and catching!

Let me test your change on Windows and run unit tests before submission.

CMakeLists.txt
```diff
@@ -131,6 +131,11 @@ install(
     PATTERN "openmp_wrappers" EXCLUDE)
 
 option(FILE_REORG_BACKWARD_COMPATIBILITY "Enable File Reorg with backward compatibility" ON)
+#get rid of any RPATH definations already
+set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "")
+#set RPATH for the binary
+set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--disable-new-dtags -Wl,--rpath,${CPACK_PACKAGING_INSTALL_PREFIX}/lib" )
```
emankov: 1. This code is for linking by GNU linker only. Both options are neither supported by the Microsoft C/C++ link tool nor have analogues. Suggestion: hide the above code under `if(UNIX)` condition.
2. Setting `--rpath` to `${CPACK_PACKAGING_INSTALL_PREFIX}/lib` assumes the packaging. We need to cover the cases when hipify-clang is built (and installed/not installed by cmake), and not used from the package installed. We need to cover those cases because hipify-clang is actually an independent standalone open-source tool that might be used separately from ROCm HIP.
3. Btw, unit-testing with your PR is ok on Windows 10 (if (1) is fixed) and Ubuntu 21.10.

emankov (changes requested):

emankov (approved):
Thanks! The below rpath suits now both packaged installation and just building, and only targets `UNIX`:

```cmd
0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/../lib]
```

Thanks for catching that.

LGTM!

PR #690, opened 2022-11-11
[HIPIFY][SWDEV-310152][package][build][fix] Remove rpath from binary files

raramakr:
SWDEV-310152:
- Single version package: add ldconfig, no RPATH/RUNPATH in either binaries or libraries 
- Multi version package: use RPATH for binaries, no RPATH/RUNPATH for libraries 
- amdclang/clang/hipcc compilers shall not add RPATH/RUNPATH to produced binaries or libraries unless explicitly requested by an option

Versioning scripts will take care of adding rpath to binaries for multi version package


emankov (approved):
LGTM!

PR #775, opened 2023-02-16
[HIPIFY][#690][#771][SWDEV-310152][package][build][5.5.0] Use RUNPATH rather than using RPATH

raramakr:


CMakeLists.txt
```diff
@@ -153,7 +153,7 @@ if(UNIX)
     #get rid of any RPATH definations already
     set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "")
     #set RPATH for the binary
-    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--disable-new-dtags -Wl,--rpath,$ORIGIN/../lib" )
+    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,$ORIGIN/../lib" )
```
emankov, line 156: Should this fix go with 5.5?

emankov (approved):
LGTM

CMakeLists.txt
raramakr: Since most of the ROCm libraries and binaries do have RUNPATH(which is recommended). Will bring this also to 5.5

PR #814, opened 2023-03-24
[HIPIFY][SWDEV-368821][install][fix] Add `bin` subfolder to installation destination in `CMakeLists.txt`

dipietrantonio:
Fixes the installation path of binaries.

Currently, binaries are placed in the installation prefix, at the same level as the `bin`, `lib64` directories. This fix will place binaries in the bin folder.


emankov (approved):
LGTM

PR #816, opened 2023-03-27
[HIPIFY][SWDEV-390802][build][fix] Hipify scripts and binary path updated in hipify-clang.txt

raramakr:
As part of SWDEV-368821, the scripts and binary will get installed under <Install_Prefix>/bin folder But the packaging txt file is referring the old path <Install_Prefix>. Corrected the same. 
CMAKE_INSTALL_PREFIX is used by default if the DESTINATION is a relative path. Changed the DESTINATION to relative path


CMakeLists.txt
```diff
@@ -124,18 +124,20 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
     set(CMAKE_INSTALL_PREFIX "${PROJECT_BINARY_DIR}/dist" CACHE PATH ${INSTALL_PATH_DOC_STRING} FORCE)
 endif()
 
-set(HIPIFY_INSTALL_PATH ${CMAKE_INSTALL_PREFIX})
+set(HIPIFY_BIN_INSTALL_PATH "${CMAKE_INSTALL_PREFIX}/bin")
 
-install(TARGETS hipify-clang DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
+install(TARGETS hipify-clang DESTINATION bin)
```
emankov, line 129: Shouldn't it be 
```cmake
install(TARGETS hipify-clang DESTINATION $(HIPIFY_BIN_INSTALL_PATH))
```
instead?

raramakr: By default ${CMAKE_INSTALL_PREFIX} will be added to install command DESTINATION, if the path is relative(ex: bin)
Better to use relative path in destination , rather than absolute path
https://cmake.org/cmake/help/latest/command/install.html

emankov (approved):
LGTM, thanks!

PR #937, opened 2023-07-05
[HIPIFY][SWDEV-379066][build] Add the ability to include hipify in HIP SDK

kzhuravl:
Add the ability to include hipify in HIP SDK

- Enabled with -DHIPIFY_INCLUDE_IN_HIP_SDK=On
  - Must be built with LLVM_EXTERNAL_PROJECTS
    - This way it is a monolithic build, which makes all compiler and linker options consistent across the build
      - This ensures that the build will keep working if compiler and/or linker options for llvm build change
  - Must be built with DK's compiler
- Other minor cleanups

The difference between a regular build and HIP SDK build, is HIP SDK build has to be built using the compiler from the DK (which can change), and the use of ninja as a generator. If I understand correctly, the regular windows build is using Visual Studio project files, which are not supported in the HIP SDK build environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,12 +3,6 @@ cmake_minimum_required(VERSION 3.16.8)
 project(hipify-clang)
 
 include(GNUInstallDirs)
-find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
-
-message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}:")
-message(STATUS "   - CMake module path: ${LLVM_CMAKE_DIR}")
-message(STATUS "   - Include path     : ${LLVM_INCLUDE_DIRS}")
-message(STATUS "   - Binary path      : ${LLVM_TOOLS_BINARY_DIR}")
 
 option(HIPIFY_CLANG_TESTS "Build HIPIFY tests, if lit is installed" OFF)
 option(HIPIFY_CLANG_TESTS_ONLY "Build HIPIFY tests only, if lit is installed and hipify-clang binary is already produced" OFF)
@@ -31,8 +25,51 @@ file(GLOB_RECURSE HIPIFY_HEADERS src/*.h)
 add_llvm_executable(hipify-clang ${HIPIFY_SOURCES} ${HIPIFY_HEADERS})
 target_link_directories(hipify-clang PRIVATE ${LLVM_LIBRARY_DIRS})
 
-set(CMAKE_CXX_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang++)
-set(CMAKE_C_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang)
+set(HIPIFY_INSTALL_ROOT_DIR ${CMAKE_INSTALL_PREFIX})
+set(HIPIFY_INSTALL_BIN_DIR ${HIPIFY_INSTALL_ROOT_DIR/bin})
+
+set(INSTALL_PATH_DOC_STRING "hipify-clang Installation Path")
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    set(CMAKE_INSTALL_PREFIX "${PROJECT_BINARY_DIR}/dist" CACHE PATH ${INSTALL_PATH_DOC_STRING} FORCE)
+endif()
+
+option(HIPIFY_INCLUDE_IN_HIP_SDK "Include HIPIFY in HIP SDK" OFF)
+if(HIPIFY_INCLUDE_IN_HIP_SDK)
+  if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+    message(FATAL_ERROR "In order to include HIPIFY in HIP SDK, HIPIFY needs to be built with LLVM_EXTERNAL_PROJECTS")
+  endif()
+
+  # Need to add clang include directories explicitly if
+  # building as part of llvm.
+  if(LLVM_EXTERNAL_CLANG_SOURCE_DIR)
+    target_include_directories(hipify-clang
+      PRIVATE
+        ${LLVM_BINARY_DIR}/tools/clang/include
+        ${LLVM_EXTERNAL_CLANG_SOURCE_DIR}/include)
+  endif()
+  # Need to add lld include directories explicitly if
+  # building as part of llvm.
+  if(LLVM_EXTERNAL_LLD_SOURCE_DIR)
+    target_include_directories(hipify-clang
+      PRIVATE
+        ${LLVM_BINARY_DIR}/tools/lld/include
+        ${LLVM_EXTERNAL_LLD_SOURCE_DIR}/include)
+  endif()
+
+  set(CLANG_RESOURCE_HEADERS ${CMAKE_INSTALL_PREFIX}/lib/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH})
+else()
+  find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
+
+  message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}:")
+  message(STATUS "   - CMake module path: ${LLVM_CMAKE_DIR}")
+  message(STATUS "   - Include path     : ${LLVM_INCLUDE_DIRS}")
+  message(STATUS "   - Binary path      : ${LLVM_TOOLS_BINARY_DIR}")
+
+  set(CMAKE_C_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang)
+  set(CMAKE_CXX_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang++)
+
+  set(CLANG_RESOURCE_HEADERS ${LLVM_DIR}/../../clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH})
+endif()
 
 # Link against LLVM and CLANG libraries
 target_link_libraries(hipify-clang PRIVATE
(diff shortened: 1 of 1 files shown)


searlmc1 (approved):
LGTM

CMakeLists.txt
```diff
@@ -31,8 +25,51 @@ file(GLOB_RECURSE HIPIFY_HEADERS src/*.h)
 add_llvm_executable(hipify-clang ${HIPIFY_SOURCES} ${HIPIFY_HEADERS})
```
emankov, line 42: **[Failure]**
CMake Error at CMakeLists.txt:25 (add_llvm_executable):                                                                                                                                                                  Unknown CMake command "add_llvm_executable".

**[Reason]**
`find_package(LLVM...)` was moved below `add_llvm_executable`

CMakeLists.txt
```diff
@@ -3,12 +3,6 @@ cmake_minimum_required(VERSION 3.16.8)
 project(hipify-clang)
 
 include(GNUInstallDirs)
-find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
```
emankov, line 6: **[Failure]**
CMake Error at CMakeLists.txt:11 (include):                                                                                                                                                                              include could not find requested file:                                                                                                                                                                                                                                                                                                                                                                                                          AddLLVM

**[Reasons]**
+ `find_package(LLVM...)` was moved below `include(AddLLVM)`
+ The above `list(APPEND CMAKE_MODULE_PATH ${LLVM_CMAKE_DIR})` is also incorrect, as `LLVM_CMAKE_DIR` is yet unknown

emankov (changes requested):
The regular cmake build is broken

kzhuravl:
Fixed the issue of building hipify separately from llvm-project.

CMakeLists.txt
```diff
@@ -31,8 +35,44 @@ file(GLOB_RECURSE HIPIFY_HEADERS src/*.h)
...
-set(CMAKE_C_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang)
+set(HIPIFY_INSTALL_ROOT_DIR ${CMAKE_INSTALL_PREFIX})
+set(HIPIFY_INSTALL_BIN_DIR ${HIPIFY_INSTALL_ROOT_DIR/bin})
+
+set(INSTALL_PATH_DOC_STRING "hipify-clang Installation Path")
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    set(CMAKE_INSTALL_PREFIX "${PROJECT_BINARY_DIR}/dist" CACHE PATH ${INSTALL_PATH_DOC_STRING} FORCE)
+endif()
+
+option(HIPIFY_INCLUDE_IN_HIP_SDK "Include HIPIFY in HIP SDK" OFF)
```
emankov: `HIPIFY_INCLUDE_IN_HIP_SDK` redefinition

CMakeLists.txt
```diff
@@ -31,8 +35,44 @@ file(GLOB_RECURSE HIPIFY_HEADERS src/*.h)
...
+  # Need to add lld include directories explicitly if
+  # building as part of llvm.
+  if(LLVM_EXTERNAL_LLD_SOURCE_DIR)
+    target_include_directories(hipify-clang
+      PRIVATE
+        ${LLVM_BINARY_DIR}/tools/lld/include
+        ${LLVM_EXTERNAL_LLD_SOURCE_DIR}/include)
+  endif()
+
+  set(CLANG_RESOURCE_HEADERS ${CMAKE_INSTALL_PREFIX}/lib/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH})
```
emankov: `CLANG_RESOURCE_HEADERS` is never used if `HIPIFY_INCLUDE_IN_HIP_SDK` is set.

CMakeLists.txt
```diff
@@ -119,25 +159,22 @@ else()
...
+  DESTINATION ${HIPIFY_INSTALL_ROOT_DIR}
+  USE_SOURCE_PERMISSIONS
+  PATTERN "hipify-perl"
+  PATTERN "*.sh")
+
+# Headers are already included in HIP SDK, so skip those if including
+# HIPIFY in HIP SDK.
+if(NOT HIPIFY_INCLUDE_IN_HIP_SDK)
+  install(
+    DIRECTORY ${CLANG_RESOURCE_HEADERS}
```
emankov: 1. Why is the `if(NOT HIPIFY_INCLUDE_IN_HIP_SDK)` path changed?
2. It is incorrect to use `${CLANG_RESOURCE_HEADERS}` as a source directory here cause it might differ. Please, see the `LIB_CLANG_RES` definition depending on the LLVM version and the `D125860` definition above.

CMakeLists.txt
```diff
@@ -147,11 +184,11 @@ install(
     PATTERN "new"
     PATTERN "ppc_wrappers" EXCLUDE
     PATTERN "openmp_wrappers" EXCLUDE)
+endif()
 
 option(FILE_REORG_BACKWARD_COMPATIBILITY "Enable File Reorg with backward compatibility" ON)
 
 if(UNIX)
-
```
emankov: unneeded formating change

CMakeLists.txt
```diff
@@ -3,12 +3,16 @@ cmake_minimum_required(VERSION 3.16.8)
...
 
 include(GNUInstallDirs)
-find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
 
-message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}:")
-message(STATUS "   - CMake module path: ${LLVM_CMAKE_DIR}")
-message(STATUS "   - Include path     : ${LLVM_INCLUDE_DIRS}")
-message(STATUS "   - Binary path      : ${LLVM_TOOLS_BINARY_DIR}")
+option(HIPIFY_INCLUDE_IN_HIP_SDK "Include HIPIFY in HIP SDK" OFF)
+if(NOT HIPIFY_INCLUDE_IN_HIP_SDK)
```
emankov: Here and below, where `HIPIFY_INCLUDE_IN_HIP_SDK` is checked: shouldn't we restrict its usage to `Windows only`? I guess, we don't want UB on UNIX if `HIPIFY_INCLUDE_IN_HIP_SDK` is `ON`.

CMakeLists.txt
```diff
@@ -119,25 +159,22 @@ else()
...
+  USE_SOURCE_PERMISSIONS
+  PATTERN "hipify-perl"
+  PATTERN "*.sh")
+
+# Headers are already included in HIP SDK, so skip those if including
+# HIPIFY in HIP SDK.
+if(NOT HIPIFY_INCLUDE_IN_HIP_SDK)
+  install(
+    DIRECTORY ${CLANG_RESOURCE_HEADERS}
+    DESTINATION ${CMAKE_INSTALL_PREFIX}
```
emankov: Again, why is the `if(NOT HIPIFY_INCLUDE_IN_HIP_SDK)` path changed?

emankov (changes requested):
I don't quite understand why the default non-HIP-SDK path is changed and, moreover, so drastically.

emankov:
> * Must be built with LLVM_EXTERNAL_PROJECTS

Could you please extend the description with an explanation of why for the inclusion of hipify-clang into HIP-SDK on Windows, building `hipify-clang` with `LLVM_EXTERNAL_PROJECTS` is a must? And what are the main differences from a regular building of hipify-clang for Windows?

CMakeLists.txt
kzhuravl: Sure.

kzhuravl:
> > * Must be built with LLVM_EXTERNAL_PROJECTS
> 
> Could you please extend the description with an explanation of why for the inclusion of hipify-clang into HIP-SDK on Windows, building `hipify-clang` with `LLVM_EXTERNAL_PROJECTS` is a must? And what are the main differences from a regular building of hipify-clang for Windows?

Updated the description.

CMakeLists.txt
```diff
@@ -3,12 +3,23 @@ cmake_minimum_required(VERSION 3.16.8)
...
-message(STATUS "   - Binary path      : ${LLVM_TOOLS_BINARY_DIR}")
+option(HIPIFY_INCLUDE_IN_HIP_SDK "Include HIPIFY in HIP SDK" OFF)
+
+if(HIPIFY_INCLUDE_IN_HIP_SDK)
+  if(NOT WIN32)
+    message(FATAL_ERROR "HIPIFY_INCLUDE_IN_HIP_SDK is only supported on Windows")
+  endif()
+endif()
+
+if(NOT HIPIFY_INCLUDE_IN_HIP_SDK)
```
emankov: Now the below `include(AddLLVM)` doesn't work if `HIPIFY_INCLUDE_IN_HIP_SDK` is on.

CMakeLists.txt
```diff
@@ -31,8 +42,39 @@ file(GLOB_RECURSE HIPIFY_HEADERS src/*.h)
 add_llvm_executable(hipify-clang ${HIPIFY_SOURCES} ${HIPIFY_HEADERS})
 target_link_directories(hipify-clang PRIVATE ${LLVM_LIBRARY_DIRS})
 
-set(CMAKE_CXX_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang++)
-set(CMAKE_C_COMPILER ${LLVM_TOOLS_BINARY_DIR}/clang)
+set(HIPIFY_INSTALL_ROOT_DIR ${CMAKE_INSTALL_PREFIX})
+set(HIPIFY_INSTALL_BIN_DIR ${HIPIFY_INSTALL_ROOT_DIR/bin})
```
emankov: I still do not understand why the existing path is changed and/or refactored. I believe it should be left untouched. Then, we can refactor it, if needed.

CMakeLists.txt
```diff
@@ -3,12 +3,23 @@ cmake_minimum_required(VERSION 3.16.8)
...
-find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
 
-message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}:")
-message(STATUS "   - CMake module path: ${LLVM_CMAKE_DIR}")
-message(STATUS "   - Include path     : ${LLVM_INCLUDE_DIRS}")
-message(STATUS "   - Binary path      : ${LLVM_TOOLS_BINARY_DIR}")
+option(HIPIFY_INCLUDE_IN_HIP_SDK "Include HIPIFY in HIP SDK" OFF)
+
+if(HIPIFY_INCLUDE_IN_HIP_SDK)
+  if(NOT WIN32)
```
emankov, line 9: It doesn't exclude `MSVC`. Should we strict it more (like below) or do we want to support also `-G "Visual Studio XX 20YY"` for `HIPIFY_INCLUDE_IN_HIP_SDK` set to `ON`?

```cmake
if(HIPIFY_INCLUDE_IN_HIP_SDK)
  if(NOT WIN32)
    message(FATAL_ERROR "HIPIFY_INCLUDE_IN_HIP_SDK is only supported on Windows")
  elseif(MSVC)
    message(FATAL_ERROR "HIPIFY_INCLUDE_IN_HIP_SDK is targeting Ninja only")
  endif()
endif()
```

emankov (changes requested):

CMakeLists.txt
kzhuravl: MSVC is a compiler, not a build system, your code snippet above will always hit the "HIPIFY_INCLUDE_IN_HIP_SDK is targeting Ninja only" error with HIPIFY_INCLUDE_IN_HIP_SDK=ON, because we have to use MSVC from the DK.

I can restrict it to Ninja, but I am not sure if it is the right choice. For example, someone could use regular makefiles, which would work fine on windows.

I think it would be better to check that it is not Visual Studio, like in the following code snippet:

```
if(CMAKE_GENERATOR MATCHES "Visual Studio")
  message(FATAL_ERRROR "...")
endif()
```

CMakeLists.txt
kzhuravl: That is not true. I am doing monolithic build, so that works fine.

CMakeLists.txt
kzhuravl: This is not true. I am doing monolithic build, so that works fine.

CMakeLists.txt
kzhuravl: I have dropped all of the cleanups and refactoring and left the bare minimum to make it "includable" in the HIP SDK. Feel free to do those cleanups and refactoring later.

I had to change the **install** command for hipify-perl:

`DIRECTORY ${CMAKE_SOURCE_DIR}/bin -> DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/bin`

it works "as is" if you build hipify standalone, but it is incorrect to use CMAKE_SOURCE_DIR if you build it as part of other project, because CMAKE_SOURCE_DIR may point to other project, however CMAKE_CURRENT_SOURCE_DIR will point to the current source directory...

kzhuravl:
I have tested a standalone build on linux (regular cmake build) - it works fine. I have tested on windows (as part of HIP SDK) - it also works fine.

CMakeLists.txt
emankov: That's better and enough, I think. Thanks!

emankov:
> I have tested a standalone build on linux (regular cmake build) - it works fine. I have tested on windows (as part of HIP SDK) - it also works fine.

So did I, except for the case with HIP SDK. I did both Win and Lin standalone builds with and without testing. Everything looks fine now.

emankov (approved):
LGTM, Thank you!

CMakeLists.txt
emankov: I meant the case when `HIPIFY_INCLUDE_IN_HIP_SDK` is ON but not in the monolithic build. I'll exclude this case in a separate change.

PR #962, opened 2023-07-19
[HIPIFY][SWDEV-409236][perl] - Get the absolute path of the script directory

raramakr:
Softlinks of the hipify scripts will be there in /usr/bin directory due to update-alternatives execution. Invoking scripts from softlinks results in failure, since the script directory path was incorrect. Use realpath to get the absolute paths

### DIFF
--- a/bin/hipconvertinplace.sh
+++ b/bin/hipconvertinplace.sh
@@ -6,7 +6,7 @@
 # This can be quite handy when dealing with an existing CUDA code base since the script
 # preserves the existing directory structure.
 
-SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+SCRIPT_DIR="$(dirname "$(realpath "$0")")"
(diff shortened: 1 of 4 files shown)


bin/hipconvertinplace.sh
```diff
@@ -6,7 +6,7 @@
 # This can be quite handy when dealing with an existing CUDA code base since the script
 # preserves the existing directory structure.
 
-SCRIPT_DIR=`dirname $0`
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
```
amd-isparry: The manual page for readlink suggests using realpath is preferred.

raramakr: Thanks Icarus. Updated the patch by using realpath for getting the absolute path

emankov (approved):
LGTM, thanks!

PR #997, opened 2023-08-22
[HIPIFY][SWDEV-356700][doc] Mention and link to hipify_torch

jeffdaily:


emankov (approved):
LGTM. Btw, what about auto-generating of hipify_torch source code by hipify-clang as it's already done for hipify-perl? It looks like hipify_torch's synchronization with hipify-clang sources and with the CUDA2HIP auto-generated documentation is not automated.

PR #1033, opened 2023-09-20
[HIPIFY][doc] Replace build.image with build.os in the .readthedocs.yaml config

samjwu:
https://docs.readthedocs.io/en/stable/config-file/v2.html

build.image (which is implicitly used if neither is specified) will be deprecated in October 2023


emankov (approved):
LGTM

PR #1177, opened 2023-12-01
[HIPIFY][doc] Fix links in Sphinx table of contents

samjwu:
Context: Sphinx does not parse anchor tags correctly, resulting in an offset in the Sphinx table of contents links (left nav bar): https://rocm.docs.amd.com/projects/HIPIFY/en/latest/

Change: This fixes it by using pure Markdown for the header, which does not break any links in the README

Checked that links in the README.md still work: https://github.com/samjwu/HIPIFY/blob/doc-toc/README.md


emankov (approved):
LGTM, thanks!

PR #1239, opened 2023-12-28
[HIPIFY][cmake] Allow custom LLVM for build from src

djtodoro:
Introduce CUSTOM_CLANG_PATH cmake variable that can be used when building HIPIFY from source.
Used as following:

```
$ cmake -DCUSTOM_CLANG_PATH=${path_to_custom_llvm_build}/lib/cmake/clang ...
```


djtodoro:
cc @emankov

CMakeLists.txt
```diff
@@ -32,8 +32,13 @@ if(HIPIFY_INCLUDE_IN_HIP_SDK)
...
 else()
-  find_package(Clang REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
-  find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
+  if (CUSTOM_CLANG_PATH)
+    find_package(Clang REQUIRED CONFIG PATHS ${CUSTOM_CLANG_PATH} NO_DEFAULT_PATH)
+    message(STATUS "Using ClangConfig.cmake in: ${CUSTOM_CLANG_PATH}")
+  else()
+    find_package(Clang REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
+    find_package(LLVM REQUIRED CONFIG PATHS ${CMAKE_PREFIX_PATH})
+  endif()
```
emankov, line 41: `CMAKE_PREFIX_PATH` is already used for specifying `custom` LLVM/clang. What are the reasons for introducing `CUSTOM_CLANG_PATH` for a separate search for the `Clang` package only?

emankov (changes requested):
The title `Allow custom LLVM for build from src` misleads, as only the `Clang` package is being searched, in fact.

CMakeLists.txt
djtodoro: It will automatically pick up llvm.cmake configuration when you specify the path to clang's one, but yeah, it was not clear to me from the doc that `CMAKE_PREFIX_PATH` is being used for that purposes only. If that is the case, I will drop this. Thanks :)

emankov: I'll make custom LLVM related text more clear in the README.md. Thanks.

djtodoro: Nice! Thanks!

PR #1276, opened 2024-01-10
[HIPIFY][SWDEV-435473][cmake][fix] The installation location of hipify_clang is not correct

aakanksha555:
Make the cmake install and package install locations consistent.


emankov (approved):
LGTM, Thanks!

PR #1380, opened 2024-02-26
[HIPIFY][docs] Updated the usage of `hipify-clang`

emankov:
+ Added the `Hipification Statistics` section
+ Mentioned explicitly multiple source files support


docs/hipify-clang.md
```diff
@@ -213,23 +213,99 @@ For example:
...
+
 For some hipification automation (starting from clang 8.0.0), it is also possible to provide a [Compilation Database in JSON format](https://clang.llvm.org/docs/JSONCompilationDatabase.html) in the `compile_commands.json` file:
 
 ```bash
 -p <folder containing compile_commands.json> or
 -p=<folder containing compile_commands.json>
 ```
 
-The compilation database should be provided in the `compile_commands.json` file or generated by clang based on CMake; options separator `'--'` must not be used.
+The compilation database should be provided in the `compile_commands.json` file or generated by clang based on CMake; multiple source files can be specified as well; options separator `'--'` must not be used.
```
Rmalavally: Recommended changes,

Add the following sentence as a note, as it seems crucial. Also, what happens if users accidentally use it? What is the recommended separator? 

**Note**: options separator `'--'` must not be used.

emankov: > Recommended changes,
> 
> Add the following sentence as a note, as it seems crucial. Also, what happens if users accidentally use it? What is the recommended separator?

A compilation error will occur due to the hipify-only options provided before the separator. Unfortunately, the Compilation Database is an entirely clang-based solution, and it doesn't understand specific options of clang-tooling-based tools like hipify-clang. 
 
> **Note**: options separator `'--'` must not be used.

What about the following rewording?

The options provided in the `compile_commands.json` should contain only clang options. hipify-clang options can be provided only in the hipify-clang command line.

Rmalavally: Thanks, Evigeny. From a user perspective, informing them about the consequences of using the incorrect options separator is recommended.. 

How about

Only Clang options must be provided in compile_commands.json. hipify-clang options can be provided in the hipify-clang command line.

**Note**: Do not use the options separator '--'. A compilation error will occur if the hipify-only options are provided before the separator.

emankov: What about adding the following:

Here is an example of the` compile_commands.json` usage: https://github.com/ROCm/HIPIFY/blob/amd-staging/tests/unit_tests/compilation_database/compile_commands.json.in

Rmalavally: Perfect for additional clarity. Examples always help. So, can we revise the original description to,

Only Clang options must be provided in compile_commands.json. hipify-clang options can be provided in the hipify-clang command line. You may refer to the following *example link* for more details. 

Note: Do not use the options separator '--'. A compilation error will occur if the hipify-only options are provided before the separator.

emankov: Done

Rmalavally (approved):
Perfect. Thanks, Evigeny!

PR #1381, opened 2024-02-26
[HIPIFY][docs] Formatting

emankov:


docs/hipify-clang.md
```diff
@@ -7,7 +7,7 @@ After applying all the matchers, the output HIP source is produced.
 **Advantages:**
 
 1. It is a translator; thus, any even very complicated constructs will be parsed successfully, or an error will be reported.
```
Rmalavally, line 9: Recommended

>It is a translator; thus, any even very complicated constructs will be parsed successfully, or an error will be reported.

1. `hipify-clang` is a translator; any complex constructs will be parsed successfully, or an error will be reported.

docs/hipify-clang.md
```diff
@@ -7,7 +7,7 @@ After applying all the matchers, the output HIP source is produced.
 **Advantages:**
 
 1. It is a translator; thus, any even very complicated constructs will be parsed successfully, or an error will be reported.
-2. It supports clang options like [`-I`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-i-dir), [`-D`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-d-macro), [`--cuda-path`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-cuda-path), etc.
+2. It supports `clang` options like [`-I`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-i-dir), [`-D`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-d-macro), [`--cuda-path`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-cuda-path), etc.
 3. Seamless support of new CUDA versions as it is the clang's responsibility.
```
Rmalavally, line 11: Recommeded,

>Seamless support of new CUDA versions as it is the clang's responsibility.

To

Seamless support of new CUDA versions, as it is the clang's responsibility.

Do we mean 'hipify-clang' or Clang? Why is it clang's responsibility? Have we provided this information to users somewhere? Can we add a link for context?

emankov: Here in the sentence, `clang` is meant. Clang's responsibility is to support different CUDA versions. We already provided here above the link to [`--cuda-path`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-cuda-path), where CUDA of different versions might be specified.

emankov: Btw, what is preferable to use in the docs `Clang` or `clang`?

docs/hipify-clang.md
emankov: I will consider your recommendations in the upcoming separate doc changes. This one is about formatting only.

docs/hipify-clang.md
Rmalavally: From what I understand, the generic use is Clang (https://clang.llvm.org/). Usage depends on the context. I have seen 'hipify-clang' used consistently, so we may be good to keep that usage. 

Based on your use of 'clang', are you referring to Clang from LLVM or 'hipify-clang' from context?

emankov: > Based on your use of 'clang', are you referring to Clang from LLVM or 'hipify-clang' from context?

`hipify-clang` contains `clang` statically linked inside; at least those `clang` and `LLVM` libs that are needed for `hipify-clang`. Saying about something that it is `the clang's responsibility`, the `clang` statically linked into `hipify-clang` is actually meant. Probably, we need to point it out somehow and somewhere in `hipify-calng` documentation. It is obvious only from the `CMakeLists.txt` file, which is not the documentation.

Rmalavally: You are correct. It would be good to explain why and how it is Clang's responsibility. Is it safe to assume that our users are aware of 'hipify-clang' containing clang? (I know we have clang in the name :) But new/novice users. 

We could consider providing a brief description about CMakeLists.txt, and how it connects to this 'responsibility'.

emankov: OK, I'll try to find some words for that in a separate change.

PR #1382, opened 2024-02-27
[HIPIFY][fix] Add exclusions for copying openmp headers.

estewart08:
Tensorflow is failing due to broken openmp header
links. These should not be copied in for hipify.


emankov (approved):
LGTM, thanks!

PR #1387, opened 2024-03-01
[HIPIFY][package][fix] Use the debug package creation variables provided by build script

raramakr:
SWDEV-428387 -The debug package creation will strip the binaries and thereby removing hard coded paths in binaries and libraries.
hipify-clang binary was not stripped and having hard coded ROCm path. Enabling debug package will remove the same


emankov (approved):
LGTM, thank you!

PR #1410, opened 2024-03-21
[HIPIFY][doc] MD -> RST migration - Step 1

LisaDelaney:
+ table revamp


README.md
```diff
@@ -3,28 +3,6 @@
 HIPIFY is a set of tools that you can use to automatically translate CUDA source code into portable
 [HIP](https://github.com/ROCm/HIP) C++.
 
-## Table of contents
```
emankov, line 6: Why was the table of contents removed?

docs/hipify-clang.md
```diff
@@ -1,706 +0,0 @@
-# hipify-clang
```
emankov, line 1: It is better to rename the file in one change and then make the changes in the renamed one; otherwise, it is almost impossible to review.

emankov (commented):
1. What about bold font? Is it available in the RST? 
2. What about multiple words (sentences) in a single line of the table cell?

README.md
Rmalavally: @LisaDelaney Please confirm.

LisaDelaney: We've already linked to the docs so there's no real need for a TOC in the readme...but mostly I removed this because the readme is in markdown and the file that was linked to is now RST, which makes the linking messier than it's worth dealing with (I'm slowly removing TOCs/extended content from readmes in favor of pointing user to the docs website)

docs/hipify-clang.md
LisaDelaney: the markdown file was deleted (I kept a working copy of that file/content while I was building the RST file, which replaces the markdown version)

README.md
LisaDelaney: We've already linked to the docs so there's no real need for a TOC in the readme...but mostly I removed this because the readme is in markdown and the file that was linked to is now RST, which makes the linking messier than it's worth dealing with (I'm slowly removing TOCs/extended content from readmes in favor of pointing user to the docs website)

randyh62 (approved):

docs/hipify-clang.md
emankov: It is not quite so. First, the file is renamed. Second, the content of the file is changed. If it is in two different commits or PRs, the second allows us to see the diff. Could you please try it on the rest of the md files with your md -> rest migration next time?

emankov (approved):
LGTM, thanks!

PR #1495, opened 2024-05-21
Update RTD config to use Python 3.10 and rocm-docs-core 1.1.2

samjwu:
Update RTD config to use Python 3.10 and rocm-docs-core 1.1.2

Summary of proposed changes:
- ReadtheDocs will use Python 3.10 instead of Python 3.8 to build docs
  - To use newer packages (eg: Sphinx 7)
- Doc builds will use rocm-docs-core version 1.1.2


emankov (approved):
LGTM

PR #1613, opened 2024-08-09
[HIPIFY][fix] Header device_launch_parameters present in CUDA should be skipped

cjatin:
Initially mentioned by https://github.com/ROCm/HIP/issues/3568

the header device_launch_parameters.h is not present in HIP, it should be consumed when its encountered during hipify process.

### DIFF
--- a/bin/hipify-perl
+++ b/bin/hipify-perl
@@ -5635,6 +5635,7 @@ sub simpleSubstitutions {
     subst("cublas.h", "hipblas.h", "include_cuda_main_header");
     subst("cuda.h", "hip\/hip_runtime.h", "include_cuda_main_header");
     subst("cuda_runtime.h", "hip\/hip_runtime.h", "include_cuda_main_header");
+    subst("device_launch_parameters.h", "", "include");
     subst("cudnn.h", "hipDNN.h", "include_cuda_main_header");
     subst("cufft.h", "hipfft\/hipfft.h", "include_cuda_main_header");
     subst("curand.h", "hiprand\/hiprand.h", "include_cuda_main_header");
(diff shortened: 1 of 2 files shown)


emankov:
Hello @cjatin,

What about declarations and definitions presented in `device_launch_parameters.h`? 
Are they all presented in HIP and also reachable w/o explicit header file inclusion? 
For instance, `threadIdx`, `blockIdx`, etc. look reachable implicitly through `#include <hip/hip_runtime.h>`.

bin/hipify-perl
```diff
@@ -5635,6 +5635,7 @@ sub simpleSubstitutions {
     subst("cublas.h", "hipblas.h", "include_cuda_main_header");
     subst("cuda.h", "hip\/hip_runtime.h", "include_cuda_main_header");
     subst("cuda_runtime.h", "hip\/hip_runtime.h", "include_cuda_main_header");
+    subst("device_launch_parameters.h", "", "include");
```
emankov: Have you been guided by https://github.com/ROCm/HIPIFY/blob/amd-staging/bin/hipify-perl#L25:

`# IMPORTANT: Do not change this file manually: it is generated by hipify-clang --perl`

and regenerated hipify-perl, or this is a manual change in `hipify-perl`?

cjatin:
I think so. From what I know device_launch_parameters contains thread/block/grid id and dims and warpSize. which are all reachable from hip/hip_runtime.h

emankov:
> I think so. From what I know device_launch_parameters contains thread/block/grid id and dims and warpSize. which are all reachable from hip/hip_runtime.h

We need to inspect the rest. Those that are not reachable need explicit inclusion of the corresponding header; those that are not supported need to be marked as HIP_UNSUPPORTED.

And, btw, what about hipify-perl? Was it changed manually or regenerated?

cjatin:
> And, btw, what about hipify-perl? Was it changed manually or regenerated?

My bad, I manually edited it. Missed the `IMPORTANT: Do not change this file manually: it is generated by hipify-clang --perl` on the top of the file. Fixed it

emankov (approved):
LGTM,

But we need to inspect the rest of the data types. Those that are not reachable need explicit inclusion of the corresponding header; those that are not supported need to be marked as HIP_UNSUPPORTED [separated change].

PR #1662, opened 2024-09-23
[HIPIFY][doc] Updated the information for 6.2.1 in the changelog

spolifroni-amd:
Updated the changelog file for 6.2.1 so that it reflects what is in the 6.2.1 Release Notes and conforms to the documentation standards. 

The changelog for subsequent versions should conform to this standard.


emankov (changes requested):
Version `18.0.0` is incorrect; the correct version is 6.2.1.

If you want to add an "additional" word `Added` everywhere, it should be done for the whole Changelog, but I'm against it.

FMPOV, this change has to be abandoned.

emankov:
Abandoned in favour of the #1717

PR #1808, opened 2024-12-20
[HIPIFY][doc] `LLVM 19.1.6` is the latest supported LLVM release

emankov:
+ No patches are needed
+ Updated the `README.md` accordingly
+ `hipify-clang` built with `LLVM 19.1.6` works correctly with the latest supported `CUDA 12.6.3`, even though clang may report that `CUDA 12.6.3` is not fully supported
+ Tested on `Windows 11` (`VS 2019` and `VS 2022`)  and `Ubuntu 23.10`


kunal-mansukhani:
```
./hipify-clang  --version
Ubuntu LLVM version 19.1.6
  Optimized build.
munmap_chunk(): invalid pointer
pure virtual method called
terminate called without an active exception
Aborted
```

I'm getting this error on LLVM v19.1.6

PR #1840, opened 2025-01-22
[HIPIFY][build][clang][fix] Make hipify-clang buildable with an LLVM configured to link `libLLVM.so`

stellaraccident:
Without these changes, if building against an LLVM configured with `-DLLVM_LINK_LLVM_DYLIB=ON`, the resulting binary will have ODR violations resulting in the dread duplicate CL option registered errors at runtime.

In order to work in this case, executables must be uniformly linked in the same way that LLVM binaries themselves are: by setting `LLVM_LINK_COMPONENTS` vs depending directly on the (static) individual libraries. This should work in all linking modes.

Note that the problem originate because an LLVM configured in this way will configure any clang libraries to carry a transitive dep on libLLVM. Therefore, it is illegal to depend on clang (static) libraries and LLVM static libraries in combination.

There is a step further that could be taken to also support building against a shared libClang, but this is merely an optimization vs an error: since libClang is a leaf library, it is ok to depend on its static components. Perhaps total package size could be reduced by also supporting libClang, but this is left for the future.


emankov:
Hi @stellaraccident,

First, thanks for your contribution to the project.
Second, we need some more time to check your fix.

Thanks!

searlmc1:
Hi @stellaraccident ,

Posting the same comment from https://github.com/ROCm/HIPIFY/pull/1841 - is it critical to land this in the ROCm:rocm-6.3.x branch ? If so, we need to coordinate with ROCm-devOps as the source-of-truth for that branch is internal. Also, I assume this should land in the amd-staging branch. E.g., I assume this issue also exists in amd-staging / also needs to be fixed there.

Thanks much

jayhawk-commits:
Hello @searlmc1 , we changed the target branch to amd-staging on behalf of @stellaraccident  to go through the usual flow to eventually get the change landed to mainline. The fork was rebased to latest amd-staging HIPIFY but it's picked up these extra commits from mainline that can be cleaned up with Squash and Merge.

emankov:
Thank you @amd-jmacaran, but I think with change will be abandoned. There will be another change for using `libLLVM.(so | lib)`, which needs more testing for multiple OSs. No estimate due to doing it in the background.

emankov:
Closing this PR in favor of #1873

Issue #1790, opened 2024-12-12
[HIPIFY][build] Switch `CPACK_PACKAGE_VERSION` from `LLVM` to `ROCm`


emankov:
### Problem Description

The HIPIFY version in Linux packages should be the same as the ROCm version, not the LLVM version.

### Operating System

any

### CPU

any

### ROCm Version

ROCm 6.0.0

### ROCm Component

HIPIFY

kzhuravl:
PR: https://github.com/ROCm/HIPIFY/pull/1867

PR #1867, opened 2025-02-26
[HIPIFY][#1790][build] Switch CPACK_PACKAGE_VERSION from LLVM to ROCm

aakanksha555:


CMakeLists.txt
```diff
@@ -51,6 +51,30 @@ endif()
 list(APPEND CMAKE_MODULE_PATH ${LLVM_CMAKE_DIR})
 include(AddLLVM)
 
+# Extract ROCM version from HIPIFY sources
+file(READ src/Statistics.h CONTENTS)
```
emankov: To obtain the ROCm version from HIPIFY itself, we can set it directly in the CMakeLists.txt—I don't think we need such complications as obtaining it from sources. 

I'm wondering about obtaining the ROCm version from above when hipify-clang building/packaging is performed in the ROCm packaging harness. What do you think about such a version detection as a first step?

By the way, the minor (or patch, or the third-digit) ROCm version in HIPIFY sources is not always correct, as not every minor release of HIPIFY tools has changes, so the minor version is 0 in most cases. That is also why the ROCm version detection from ROCm itself looks more preferable. Moreover, we need ROCm version instead of LLVM version for the correct hipify package file naming, first of all, and it should be exactly the same as the versions of the rest of the ROCm packages.

emankov (changes requested):

emankov:
Hello @aakanksha555, do we have a chance to fix it in 6.5.0 (soon)?

aakanksha555:
Sorry, I inadvertently closed this PR. I just wanted a clean amd-staging branch to work off of. 
This work is at - https://github.com/aakanksha555/HIPIFY/tree/paakan/cpack_package_version

emankov:
> Sorry, I inadvertently closed this PR. I just wanted a clean amd-staging branch to work off of. This work is at - https://github.com/aakanksha555/HIPIFY/tree/paakan/cpack_package_version

It should be reworked anyway. We need another solution, as I tried to explain it above.

PR #1973, opened 2025-05-28
[HIPIFY][#SWDEV-534480][BLAS] Support for hipBLAS 7.0 - Step 2 - renaming `v2` APIs - Part 2 - `hipblasGemm(Strided)?(Batched)?Ex_v2(_64)?`

emankov:
+ [IMP] `hipBLAS 7.0` doesn't support the following API in CUDA ABI < 11.0 anymore:
  - `cublasGemmEx`
  - `cublasGemmBatchedEx`
  - `cublasGemmStridedBatchedEx` 
 
   As the analogous `hipBLAS` API `hipblasGemmEx`, `hipblasGemmBatchedEx`, and `hipblasGemmStridedBatchedEx` changed their ABI to the same ABI as the above CUDA API, which got it only starting with CUDA 11.0: in these APIs, `hipBLAS` now uses `hipblasComputeType_t` instead of `hipDataType`
+ Updated the `hipBLAS` tests, the regenerated `hipify-perl` and `BLAS` `CUDA2HIP` docs accordingly
+ [ToDo] Rename the rest `_v2` functions


TorreZuk (commented):
Looks good to me

PR #2090, opened 2025-09-02
[HIPIFY][feature][SWDEV-541206] Introduces implicit header hipification support

ranapratap55:
### Purpose
Introducing optional local header hipification to hipify-clang, enabling automatic CUDA-to-HIP translation headers that are referenced via quoted includes (`#include "sample.h"`).

### Problem Statement
Currently `hipify-clang` only transforms the `main.cu/cpp`. User defined local headers present in the source file are not automatically hipified, requiring developers to manually hipify all the required local headers. 

### Solution
Introduces optional pre-processing step that hipifies user defined local (quoted) headers before the main source(s). If the source file contains local headers (""), with these new CLI flags we iteratively `hipify-clang` the local header files.

### Motivation
https://github.com/NVIDIA/cuda-samples/blob/3f1c50965017932fc81e6d94a3fc9e04c105b312/Samples/5_Domain_Specific/HSOpticalFlow/main.cpp#L36
Line 35, 37, and, 38 are having local header files. Instead of doing the separate hipify-clang for each of these headers, this change implicitly hipify-clang all the local headers of the source file when we run hipification on the main.cpp

### Usage
Disabled by default. To enable the `local header hipification`  

**`--local-headers`**: Enable local header hipification (non-recursive)
**`--local-headers-recursive`**: Enable local header hipification and recursively traverse quoted includes

hipify-clang run command line changes
--------------------
Old:
No local header hipification step existed.
New:
1. Non-recursive: `hipify-clang --local-headers main.cu -I include/`
2. Recursive: `hipify-clang --local-headers-recursive main.cu -I include/`


emankov:
Hi @ranapratap55,

It will take some time to check your feature extensively. 
For now, could you please add a corresponding unit test? And add the corresponding SWDEV in the title, please.
Let me know if my assistance is needed.

And thank you!

src/LocalHeader.cpp
```diff
@@ -0,0 +1,230 @@
...
+    toProcess.pop_back();
+    if (processed.count(hdr)) continue;
+    processed.insert(hdr);
+
+    bool ok = runHipifyOnSingleFile(hdr, mainSourceAbsPath, compDB, OptionsParserPtr, hipify_exe, false);
+    if (!ok) {
+      errs() << "\n" << sHipify << sError << "Hipify failed for header: " << hdr << "\n";
+    }
+
+    if (recursive) {
```
emankov, line 155: In which scenario is a recursive hipification needed? In the change, I see only a single call with the recursive flag set to false. Also, tests for recursive usage are required.

src/ImplicitHeader.cpp
```diff
@@ -0,0 +1,230 @@
...
+                             bool recursive) {
+  std::string content;
+  if (!readFileToString(mainSourceAbsPath, content)) {
+    errs() << "\n" << sHipify << sError << "Cannot read source file: " << mainSourceAbsPath << "\n";
+    return false;
+  }
+
+  std::regex includeLineRe(R"re(^\s*#\s*include\s*"([^"]+)"\s*(?:\/\/.*)?$)re", std::regex::ECMAScript);
+
+  std::smatch m;
```
emankov: Just codding rules. If the `using namespace` is declared, a namespace prefix looks extensive. As for the HIPIFY codes, we've tried to avoid `using namespace` for better support compatibility and to minimise all possible ambiguity.

emankov:
It would be helpful also to:
1. Explain the purpose of such an implicit additional hipification (cases when it is actually needed).
2. Hide this feature under the new option.
3. Disable this new option by default.
4. [Already mentioned] Feature tests.

src/main.cpp
```diff
@@ -428,6 +430,8 @@ int main(int argc, const char **argv) {
     }
     // Initialise the statistics counters for this file.
     Statistics::setActive(src);
+    // Checks the local headers first and operates on them.
+    hipifyLocalHeaders(sSourceAbsPath, compilationDatabase.get(), &OptionsParser, argv[0], false);
```
emankov: The function is supposed to be successful/unsuccessful, isn't it?

src/LocalHeader.cpp
```diff
@@ -0,0 +1,230 @@
...
+  SmallString<256> realBuf;
+  std::error_code ec = llvm::sys::fs::real_path(in, realBuf);
+  if (!ec) return true;
+
+  SmallString<256> norm = in;
+  llvm::sys::path::remove_dots(norm, true);
+  return llvm::sys::fs::exists(norm);
+}
+
+static bool runHipifyOnSingleFile(const std::string &srcPath,
```
emankov: I believe this function should be refactored along with the corresponding `main.cpp` codes to have a solid single hipification functionality, separated into yet another function.

src/LocalHeader.cpp
```diff
@@ -0,0 +1,230 @@
...
+    std::string tmpDirAbs = getAbsoluteDirectoryPath(TemporaryDir, EC);
+    if (EC) {
+      errs() << "\n" << sHipify << sError << "Temporary dir error: " << EC.message() << "\n";
+      return false;
+    }
+    
+    SmallString<256> tmpTemplate(tmpDirAbs);
+    tmpTemplate.push_back('/');
+    tmpTemplate.append(srcFileName);
+    tmpTemplate.append(".XXXXXX.hip");
```
emankov: Not sure about such a naming approach.

emankov (changes requested):

src/LocalHeader.cpp
ranapratap55: updated the tests.

src/ImplicitHeader.cpp
ranapratap55: updated.

src/main.cpp
ranapratap55: updated with warnings/error based on the header file.

src/LocalHeader.cpp
ranapratap55: Keeping separate file for local hipification will be better for maintainability of additional features. I believe we can keep main.cpp simple.

src/LocalHeader.cpp
ranapratap55: updated.

emankov:
Still, the purpose of the feature needs to be explained in the description.

src/main.cpp
```diff
@@ -53,6 +56,16 @@ constexpr auto DEBUG_TYPE = "cuda2hip";
...
+static llvm::cl::opt<bool> OptLocalHeaders(
+  "local-headers",
+  llvm::cl::desc("Enable hipification of quoted local headers (non-recursive)"),
+  llvm::cl::init(false));
+
+static llvm::cl::opt<bool> OptLocalHeadersRecursive(
+  "local-headers-recursive",
+  llvm::cl::desc("Enable hipification of quoted local headers recursively (implies enabling local headers)"),
+  llvm::cl::init(false));
+
```
emankov: The options have to be declared in [ArgParse.cpp](https://github.com/ROCm/HIPIFY/blob/amd-develop/src/ArgParse.cpp), and also exported in [ArgParse.h](https://github.com/ROCm/HIPIFY/blob/amd-develop/src/ArgParse.h).

src/main.cpp
```diff
@@ -42,6 +42,9 @@ THE SOFTWARE.
 #include "clang/Driver/Tool.h"
 #include "clang/Frontend/TextDiagnosticPrinter.h"
 
+#include "LocalHeader.h"
+#include "llvm/Support/CommandLine.h"
```
emankov: Please remove this include after moving options to `ArgParse.cpp`.

src/LocalHeader.h
```diff
@@ -0,0 +1,42 @@
...
+#include "llvm/Support/Debug.h"
+#include "clang/Basic/Diagnostic.h"
+#include "clang/Basic/DiagnosticIDs.h"
+#include "clang/Basic/DiagnosticOptions.h"
+#include "clang/Driver/Driver.h"
+#include "clang/Driver/Compilation.h"
+#include "clang/Driver/Tool.h"
+#include "clang/Frontend/TextDiagnosticPrinter.h"
+
+namespace ct = clang::tooling;
```
emankov, line 7: Excessive

src/LocalHeader.cpp
emankov: Can't agree, as we'd still need to support the duplicated code. I'm still for refactoring to avoid code duplication (1) and achieve more structurized code (2), which would be easier to support and upstream to LLVM (3). I believe we'd have even smaller `main.cpp` after refactoring.

src/main.cpp
```diff
@@ -428,6 +441,26 @@ int main(int argc, const char **argv) {
     }
     // Initialise the statistics counters for this file.
     Statistics::setActive(src);
+    // Checks the local headers if --local-headers/--local-header-recursive specified.
+    if (OptLocalHeaders || OptLocalHeadersRecursive) {
+      bool doRecursive = OptLocalHeadersRecursive;
```
emankov: Excessive

src/main.cpp
```diff
@@ -428,6 +441,26 @@ int main(int argc, const char **argv) {
...
+                              compilationDatabase.get(),
+                              &OptionsParser,
+                              argv[0],
+                              doRecursive)) {
+        Statistics::current().hasErrors = true;
+        LLVM_DEBUG(llvm::dbgs() << "Local header hipification failed for: " << sSourceAbsPath << "\n");
+        Result = 1;
+      }
+    } else if (Verbose) {
+      llvm::outs() << sHipify
```
emankov: 1. All the existing HIPIFY messaging does not rely on Verbose mode.
2. Diagnostics should be more concrete: warning/error/message.

tests/unit_tests/headers/local_headers/mixed.h
```diff
@@ -0,0 +1,9 @@
+// RUN: %run_test hipify "%s" "%t" %hipify_args --local-headers %clang_args
+#ifndef MIXED_H
+#define MIXED_H
+// CHECK: #include <hip/hip_runtime.h>
+// CHECK-NOT: #include <cuda_runtime.h>
+#include <cuda_runtime.h>
+inline int mul(int a,int b){ return a*b; }
+inline void s(){ cudaDeviceSynchronize(); }
```
emankov: Why is the hipification of `cudaDeviceSynchronize` not checked?

tests/unit_tests/headers/local_headers/common_2.h
```diff
@@ -0,0 +1,24 @@
...
+// CHECK: void temp_1(
+// CHECK: nSolverIters,
+// CHECK: float       *v);
+// CHECK: #endif
+// CHECK-NOT: hip_runtime.h
+// CHECK-NOT: cuda_runtime.h
+#ifndef common_2_H
+#define common_2_H
+
+void temp_1(const float *I0,
```
emankov: What's the difference between `common_1.h` and `common_2.h`?
What is actually being checked, as nothing was actually hipified? Besides, not injecting include for `hip_runtime.h`.

emankov (changes requested):

ranapratap55:
> Still, the purpose of the feature needs to be explained in the description.

updated the description.

tests/unit_tests/headers/local_headers/common_2.h
ranapratap55: updated.

tests/unit_tests/headers/local_headers/mixed.h
ranapratap55: updated.

src/main.cpp
ranapratap55: updated.

src/main.cpp
ranapratap55: updated.

src/main.cpp
ranapratap55: updated.

emankov (approved):
LGTM. Will let you know the extensive testing results here.

Thank you!

PR #2155, opened 2025-10-30
[HIPIFY][tests][#2090][fix][Linux] The non-existing directory in relative path of the `#include` directive

emankov:
**[Synopsis]**
+ On Windows with MSVC, the test `relative_path.cu` passes due to the `path canonicalization`, when the relative path `relative/./sub/../common1.h` is simplified by the preprocessor before it searches for the file; thus, after preprocessor, the path is `relative/common1.h`. And because of that `path canonicalization`, the test passes.
+ On Linux with GCC 13.3, the test `relative_path.cu` fails probably due to the missing or switched off by default `path canonicalization`.

**[Solution]**
+ Add the missing `sub` directory with yet another header file to be able to submit the change.
+ [IMP] Now, the include directive after preprocessing always becomes `relative/sub/../common1.h` and the test for `relative path` passes on both OS types.

### DIFF
--- a/tests/unit_tests/headers/local_headers/relative/sub/common2.h
+++ b/tests/unit_tests/headers/local_headers/relative/sub/common2.h
@@ -1,0 +1,12 @@
+// RUN: %run_test hipify "%s" "%t" %hipify_args --local-headers-recursive %clang_args
+
+#ifndef COMMON2_H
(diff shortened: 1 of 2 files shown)


ranapratap55 (approved):
LGTM.

PR #2166, opened 2025-11-07
[HIPIFY][SWDEV-529096][fix][Linux] Updates the cuDNN and cuTENSOR version check file logic

ranapratap55:
## Motivation

`unit_tests/synthetic/libraries/cudnn2miopen_before_9000.cu` test failure observed for `cuDNN >= 9`.

## Root Cause
`cudnn2miopen_before_9000.cu` test should not run if the cuDUNN is `>= 9`. The `get_cu*_major_ver()` logic is resulting wrong path for header files`(/usr/include/include/cudnn_version_v9.h)` which results in cudnn major version is `0`. Becuase of this, the test is not excluded for `cuDNN >= 9 `versions. 

## Solution
Remove the extra `/include` from the version logic. `cmake *` command in the build steps is already including the direct header files path instead of parent path.

### DIFF
--- a/tests/lit.site.cfg.in
+++ b/tests/lit.site.cfg.in
@@ -36,18 +36,18 @@ def get_lib_major_ver(file_name):
     return 0
 
 def get_cudnn_major_ver():
-    ver_file = config.cuda_dnn_root + '/include/cudnn_version.h'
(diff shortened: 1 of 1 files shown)


emankov (approved):
LGTM, Thanks!

PR #2174, opened 2025-11-13
[HIPIFY][8.0][FILE][feature][Linux] Initial support of `cuFile` to `hipFile` hipification

derobins:
AIROCFILE-7
AIROCFILE-15

Adds types and API calls from ROCm/hipFile:
* hipfile/include/hipfile.h


src/CUDA2HIP_FILE_API_types.cpp
```diff
@@ -0,0 +1,529 @@
...
+  {"CUFILE_PARAM_SKIP_TOPOLOGY_DETECTION",                  {CUDA_129, CUDA_0, CUDA_0}},
+  {"CUFILE_PARAM_STREAM_MEMOPS_BYPASS",                     {CUDA_129, CUDA_0, CUDA_0}},
+  {"CUFILE_PARAM_LOGGING_LEVEL",                            {CUDA_129, CUDA_0, CUDA_0}},
+  {"CUFILE_PARAM_ENV_LOGFILE_PATH",                         {CUDA_129, CUDA_0, CUDA_0}},
+  {"CUFILE_PARAM_LOG_DIR",                                  {CUDA_129, CUDA_0, CUDA_0}},
+  {"IS_CUFILE_ERR",                     {CUDA_129, CUDA_0, CUDA_0}},
+  {"CUFILE_ERRSTR",                     {CUDA_129, CUDA_0, CUDA_0}},
+  {"CUDA_DRV_ERR",                      {CUDA_129, CUDA_0, CUDA_0}},
+  {"IS_CUDA_ERR",                       {CUDA_129, CUDA_0, CUDA_0}},
+};
```
derobins, line 261: We just used version placeholders, so this will probably need some fixing.

src/CUDA2HIP_Doc.cpp
```diff
@@ -125,6 +125,17 @@ namespace doc {
...
+  /* hipFile / rocFile */
+  const string sHIPFILE             = "cuFile_API_supported_by_HIP";
+  const string sHIPFILE_md          = sHIPFILE + md_ext;
+  const string sHIPFILE_csv         = sHIPFILE + csv_ext;
+  const string sHIPFILE_and_ROC_md  = sHIPFILE + sandROC + md_ext;
+  const string sHIPFILE_and_ROC_csv = sHIPFILE + sandROC + csv_ext;
+  const string sROCFILE             = "cuFile_API_supported_by_ROC";
+  const string sROCFILE_md          = sROCFILE + md_ext;
+  const string sROCFILE_csv         = sROCFILE + csv_ext;
+  const string sCUFILE              = "cuFile";
```
ranapratap55, line 132:
```suggestion
  const string sHIPFILE = "cuFile_API_supported_by_HIP";
```

Could we keep same format as before?

derobins: How is the format different? I think I just copied one of the other libraries that has both hip and roc flavors (perhaps hipSolver/rocSover) because we have both hipFile and rocFile and would like to support both.

emankov: I guess the request was to keep the existing formatting.

emankov:
@derobins, could you please let us know the first CUDA version in which [gds-nvidia-fs](https://github.com/NVIDIA/gds-nvidia-fs) (with cuFile in it) appeared in the CUDA Toolkit Linux distribution?

derobins:
> @derobins, could you please let us know the first CUDA version in which [gds-nvidia-fs](https://github.com/NVIDIA/gds-nvidia-fs) (with cuFile in it) appeared in the CUDA Toolkit Linux distribution?

It looks like cuFile was released with CUDA 11.4.0. I don't see a cuFile section in the 11.3.x docs. It's confusing, though - 11.3 and 11.2 mention the library and header file in the list of things that are included, but there's no API documentation. Maybe the answer is 11.2?

https://docs.nvidia.com/cuda/archive/11.4.0/cufile-api/index.html

We'll have to grab header files from the various CUDA releases to see how the API has evolved and when new API calls were introduced.

emankov:
@derobins, I'll detect the former versions in the [fs releases](https://github.com/NVIDIA/gds-nvidia-fs/releases) a bit later.

src/CUDA2HIP_Doc.cpp
ranapratap55: > How is the format different? I think I just copied one of the other libraries that has both hip and roc flavors (perhaps hipSolver/rocSover) because we have both hipFile and rocFile and would like to support both.

```suggestion
  const string sHIPFILE = "cuFile_API_supported_by_HIP";
```
I meant to use existing format.

src/CUDA2HIP_FILE_API_functions.cpp
```diff
@@ -0,0 +1,183 @@
...
+  {"cuFileGetParameterSizeT",         {CUDA_0}},
+  {"cuFileGetParameterBool",          {CUDA_0}},
+  {"cuFileGetParameterString",        {CUDA_0}},
+  {"cuFileSetParameterSizeT",         {CUDA_0}},
+  {"cuFileSetParameterBool",          {CUDA_0}},
+  {"cuFileSetParameterString",        {CUDA_0}},
+};
+
+const std::map<llvm::StringRef, hipAPIChangedVersions> HIP_FILE_FUNCTION_CHANGED_VER_MAP {
+  {"hipFileHandleRegister",            {HIP_0}},
```
emankov: `{HIP_8000, HIP_0, HIP_0, HIP_8000}},, `
or
`{HIP_7000, HIP_0, HIP_0}},, `
if `hipFile` were really shipped with HIP 7.0.0

emankov (changes requested):

src/CUDA2HIP_FILE_API_functions.cpp
derobins: HIP_8000 is probably the right value, but the enum stops at HIP_7020 and I don't see any HIPIFY branches that look like they are prep for ROCm 8.0.

src/CUDA2HIP_Doc.cpp
derobins: Is this all okay now?

ranapratap55:
```suggestion
  const string sHIPFILE = "cuFile_API_supported_by_HIP";
```
still, some part of it needs to change. LGTM with a nit.

emankov (approved):
LGTM, tests, versioning, and documentation changes will come in separate changes.

**Do not upstream**

PR #2169, opened 2025-11-11
[HIPIFY][#2167][cmake][fix][DNN][tests] Fix locating version header files for `cuDNN`

emankov:
+ [Typo] The value of the `CUDNN_VERSION_FILE_NAMES` should be used: `${CUDNN_VERSION_FILE_NAMES}`
+ [ERROR] `set(CUDNN_VERSION_FILE_NAMES "cudnn_version.h cudnn_version_v9.h cudnn_version_v999.h")` creates a single string, not a list of strings

PR #2183, opened 2025-11-19
[HIPIFY][#2169][cmake][fix][tests][cuDNN][cuTENSOR] Add NO_DEFAULT_PATH and cache clearing to cuDNN/cuTensor find_file() calls

ranapratap55:
## Problem
The existing find_file() logic had two issues that caused incorrect path detection:

-  When users passed parent directories (e.g., `-DCUDA_DNN_ROOT_DIR=/usr`), find_file() would search system default paths (/usr/include) even though headers weren't in the specified location. This caused the fallback logic to be bypassed, leaving `CUDA_DNN_ROOT_DIR` pointing to the wrong location (`/usr` instead of `/usr/include`), resulting in version detection failures (returning 0 instead of 9).
- Sequential CMake runs with different paths (e.g., first with `/usr`, then `/usr/include`, then `/usr, /usr, /usr/include` of different combinations) would use cached `CUDNN_VERSION_FILE` values from previous runs, causing inconsistent and unpredictable behavior.

## Solution

- Added `NO_DEFAULT_PATH` flag to find_file() to prevent CMake from searching system default paths.
- Added `unset(CUDNN_VERSION_FILE CACHE)` and `unset(CUDA_TENSOR_FILE CACHE)` to prevent cache pollution


CMakeLists.txt
```diff
@@ -358,9 +358,11 @@ if(HIPIFY_CLANG_TESTS OR HIPIFY_CLANG_TESTS_ONLY)
         cudnn_version_v9.h
         cudnn_version_v999.h
     )
-    find_file(CUDNN_VERSION_FILE NAMES ${CUDNN_VERSION_FILE_NAMES} PATHS ${CUDA_DNN_ROOT_DIR})
+    unset(CUDNN_VERSION_FILE CACHE)
```
emankov, line 361: Does unsetting the variable to an undefined value before its first use in the `find_file` function also force CMake to perform the search and set it to the full path (if found)?

emankov:
Could you please also add a few words to the `Problem` section about the cases where we face the issue this PR fixes?

ranapratap55:
> Could you please also add a few words to the `Problem` section about the cases where we face the issue this PR fixes?

updated more info.

ranapratap55:
unsetting the variable to an undefined value before the find_file() forces CMake to perform the search from scratch.

emankov (approved):
LGTM, thanks!

PR #2191, opened 2025-11-20
[HIPIFY][7.2][RT] Sync with ROCm HIP 7.2 - Part 1 - HIP Runtime

ranapratap55:
- Added new `hipDeviceAttributeHostNumaId, hipStreamCopyAttributes(), hipLibraryEnumerateKernels(), hipKernelGetLibrary(), hipKernelGetName(), hipOccupancyAvailableDynamicSMemPerBlock()`
- Updated synthetic tests, regenerated `hipify-perl, Driver and Function CUDA2HIP` docs accordingly

PR #2193, opened 2025-11-20
[HIPIFY][#2191][tests][fix] Exclude the `runtime_functions_12080.cu` tests from running for `CUDA < 12.8`

emankov:


ranapratap55 (approved):
LGTM.

PR #2274, opened 2026-01-08
[HIPIFY][clang][compatibility][fix] Conditionally include the `<optional>` header

emankov:
+ This PR fixes the following compilation error against `LLVM < 16`:
  - Windows: `Statistics.cpp(747,86): error C2039: 'optional': is not a member of 'std'`
  - Linux: `Statistics.cpp:747:86: error: no template named 'optional' in namespace 'std'`


ranapratap55 (approved):
LGTM

PR #2292, opened 2026-01-12
[HIPIFY][build][Windows][fix] Fixing `D9025` compiler warning

emankov:
**[Synopsis]**
+ CMake adds `/EHsc` (enable exceptions) to the default flags `CMAKE_CXX_FLAGS`, which conflicts with disabling exceptions using `target_compile_options(..., /EHs-)`
+ The `cl` compiler uses the last flag it sees (`/EHs-`), but warns that the previous setting (`/EHs`) was ignored

**[Solution]**
+ Replace the existing flag in the `CMake` variable rather than just appending the new one


ranapratap55 (approved):
LGTM.

PR #2271, opened 2026-01-07
[HIPIFY][test][infra] Enable `Linux` Rock CI PSDB for `amd-mainline`

kirthana14m:

PR #2294, opened 2026-01-13
[HIPIFY][test][infra] Enable `Linux` Rock CI PSDB for `amd-staging`

kirthana14m:


skganesan008 (approved):
The commit message is too long. I am still not sure why we need to add the repository name in the prefix Evgeny. Both [HIPIFY] and [test] seems to be not needed if the intention is to differentiate between infra and non infra changes.

emankov:
> The commit message is too long. I am still not sure why we need to add the repository name in the prefix Evgeny. Both [HIPIFY] and [test] seems to be not needed if the intention is to differentiate between infra and non infra changes.

I believe I already answered in the previous PR #2271, which has been submitted to `amd-mainline` with all the suffixes.
HIPIFY lived in HIP, then it was split into its own repo, and now it is going to be upstreamed to LLVM. The prefixes must not be removed.

skganesan008:
> > The commit message is too long. I am still not sure why we need to add the repository name in the prefix Evgeny. Both [HIPIFY] and [test] seems to be not needed if the intention is to differentiate between infra and non infra changes.
> 
> I believe I already answered in the previous PR #2271, which has been submitted to `amd-mainline` with all the suffixes. HIPIFY lived in HIP, then it was split into its own repo, and now it is going to be upstreamed to LLVM. The prefixes must not be removed.

Thank you; my query/comment was about commit message getting too long. we may not be able to give a precise description of the patch at the title level sometimes with this requirement.   git commit title length recommendation seems to be around 50 chars for the title and 72 chars for the description.   It is not strictly enforced, but a standard promoted by github.  This is just something to keep in mind...

PR #2299, opened 2026-01-14
[HIPIFY][fix] `Stack Overflow` possible Issues - Part 2 - `SPARSE`, `SOLVER`

emankov:
**[Synopsis]**
+ Transformation maps reside in Stack and occupy up to `64kB` each of stack memory now
+ That leads to multiple compiler warnings about possible Stack Overflow

**[Solution]**
+ Move maps' initialization logic into a lambda. This prevents the compiler from creating a massive temporary array on the stack, which was the root cause of the "excessive stack usage" warning.


ranapratap55 (approved):
LGTM

PR #2316, opened 2026-01-21
[HIPIFY][CAFFE2] Total removal

emankov:
**[Reason]**
+ The framework `Caffe2` is deprecated. While some legacy code exists within older versions of `PyTorch`, it is not a supported path for new development. The old `rocm-caffe2` repositories you might find on GitHub (targeting ROCm 1.7 or Ubuntu 16.04) are abandoned and will not function on modern hardware (like Instinct `MI300` or `Ryzen AI`) or modern ROCm versions.


ranapratap55 (approved):
LGTM

PR #2322, opened 2026-01-26
[HIPIFY][ROCM-1367][fix] Use host target in Init() to avoid CUDA 12.x ptxas/nvlink errors

ranapratap55:
## Problem Statement
When running `hipify-clang` with CUDA 12.x, users encounter the following errors:
```
error: must pass in an explicit nvptx64 gpu architecture to 'ptxas'
error: must pass in an explicit nvptx64 gpu architecture to 'nvlink'
```
## Root Cause
The Init() function creates a Clang Driver for CUDA version detection and file sorting. Using nvptx64-nvidia-cuda target unnecessarily invokes ptxas/nvlink which require explicit --cuda-gpu-arch in CUDA 12.x. Since this compilation is only for detection/sorting (not actual device compilation), use the host target instead. 

## Solution
Change the Driver target from `nvptx64-nvidia-cuda` to the host target (`llvm::sys::getDefaultTargetTriple()`).

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -37,6 +37,11 @@ THE SOFTWARE.
 #endif
 #include "clang/Driver/Tool.h"
 #include "clang/Frontend/TextDiagnosticPrinter.h"
+#if LLVM_VERSION_MAJOR >= 16
+#include "llvm/TargetParser/Host.h"
+#else
+#include "llvm/Support/Host.h"
+#endif
 
 #include "LocalHeader.h"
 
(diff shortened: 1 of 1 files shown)


emankov (approved):
Good catch, thanks! 

With your fix, 1) `clang` still correctly adds the auxiliary triple "-aux-triple" "nvptx64-nvidia-cuda", and 2) even the warnings with the same description have been eliminated (btw, I've never faced such errors before, only warnings).

PR #2349, opened 2026-02-01
[HIPIFY][#2340][feature] Add implicit CUDA header inclusion to mimic nvcc behavior

### LINKED ISSUE #2340: [HIPIFY][#2275][feature] Implement `nvcc` behavior with `implicit CUDA headers`
**[Synopsis]**

The problem is that _**nvcc automatically pre-includes some CUDA header files at the beginning of ANY compilation unit** (not only header files, but also .cu files)_. They are at least (but probably are not all):

- **cuda_runtime.h** (The top-level injection)
- **crt/host_config.h** (System check)
- **builtin_types.h** (Basic types)
- **device_types.h** (enums and structs)
- **host_defines.h** (Macro definitions)

Such `nvcc` behaviour is a design choice by NVIDIA to make writing CUDA device and host code less verbose. It effectively treats `CUDA as part of the language environment` rather than just an external library, whereas `clang` reasonably decides that `CUDA is a C++ language extension.`

`nvcc` **knows** which header files are needed for implicit inclusion, and `clang` **doesn't**.

**[Solution]**

- Implicitly pre-include those CUDA header files during each hipification using `hipify-clang`.
- Find all the implicitly pre-included CUDA header files.
- Provide tests for all those “missing” header files.

ranapratap55:
## Purpose

Add implicit CUDA header pre-inclusion to hipify-clang to mimic nvcc's automatic header inclusion behavior.

## Problem Statement

nvcc automatically pre-includes `cuda_runtime.h` at the beginning of any CUDA compilation unit. This allows CUDA code to compile without explicit `#include` directives.

hipify-clang uses clang for parsing, which does not perform this implicit inclusion. This causes hipification to fail for some of the source files that rely on nvcc's implicit behavior.

## Solution

Introduce implicit CUDA header inclusion by adding `-include cuda_runtime.h` to clang's argument list during hipification. `cuda_runtime.h` is nvcc's single implicit entry point which transitively includes all necessary headers such as 

- crt/host_config.h (System check)
- builtin_types.h (Basic types)
- device_types.h (enums and structs)
- host_defines.h (Macro definitions)

Fixes #2340

### DIFF
--- a/src/ImplicitCudaHeaders.cpp
+++ b/src/ImplicitCudaHeaders.cpp
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2015 - present Advanced Micro Devices, Inc. All rights reserved.
+Copyright (c) 2026 - present Advanced Micro Devices, Inc. All rights reserved.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
@@ -25,24 +25,13 @@ THE SOFTWARE.
 
 namespace hipify {
 
-std::vector<std::string> getImplicitCudaHeaders() {
-  return {
-    "cuda_runtime.h"
-  };
-}
+static const char *const ImplicitCudaHeader = "cuda_runtime.h";
 
 void addImplicitCudaHeaders(ct::RefactoringTool &Tool) {
-  
-  std::vector<std::string> headers = getImplicitCudaHeaders();
-  
-  for (const auto &header : headers) {
-    Tool.appendArgumentsAdjuster(
-        ct::getInsertArgumentAdjuster(header.c_str(), 
-                                       ct::ArgumentInsertPosition::BEGIN));
-    Tool.appendArgumentsAdjuster(
-        ct::getInsertArgumentAdjuster("-include", 
-                                       ct::ArgumentInsertPosition::BEGIN));
-  }
+  Tool.appendArgumentsAdjuster(ct::getInsertArgumentAdjuster(
+      ImplicitCudaHeader, ct::ArgumentInsertPosition::BEGIN));
+  Tool.appendArgumentsAdjuster(ct::getInsertArgumentAdjuster(
+      "-include", ct::ArgumentInsertPosition::BEGIN));
 }
 
 } // namespace hipify
--- a/src/ImplicitCudaHeaders.cpp
+++ b/src/ImplicitCudaHeaders.cpp
@@ -1,0 +1,48 @@
+/*
+Copyright (c) 2015 - present Advanced Micro Devices, Inc. All rights reserved.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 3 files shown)


src/ImplicitCudaHeaders.cpp
```diff
@@ -0,0 +1,48 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+#include "ImplicitCudaHeaders.h"
+#include "clang/Tooling/ArgumentsAdjusters.h"
+
+namespace hipify {
+
+std::vector<std::string> getImplicitCudaHeaders() {
```
emankov: Should be const string(s), which are used by reference.
Actually, the function `getImplicitCudaHeaders` is excessive.

src/ImplicitCudaHeaders.h
```diff
@@ -0,0 +1,46 @@
+/*
+Copyright (c) 2015 - present Advanced Micro Devices, Inc. All rights reserved.
```
emankov: 2026

src/ImplicitCudaHeaders.cpp
ranapratap55: Updated, thanks.

src/ImplicitCudaHeaders.h
ranapratap55: updated.

emankov:
Hi @ranapratap55, 

Perhaps the last question here. Have you had a chance to try this change on the examples that were going to be "fixed" by your previous PR #2275?

ranapratap55:
> Hi @ranapratap55,
> 
> Perhaps the last question here. Have you had a chance to try this change on the examples that were going to be "fixed" by your previous PR #2275?

Hi @emankov,

Yes, I have tested this change against the examples from #2275 and changes from the #2349(this PR) alone did not fix that issue.

`CudaMath.h` Fails `helper_math.h` (float3 +=, -, * operators)
`fastWalshTransform_kernel.cuh` Fails `helper_cuda.h` (getLastCudaError macro)
`bicubicTexture_kernel.cuh` Fails `helper_math.h` (lerp, uchar)
`MonteCarlo_reduction.cuh` Fails `MonteCarlo_common.h` (project-specific __TOptionValue type)

Additionally, I also tested cuda-samples in [0_Introduction](https://github.com/NVIDIA/cuda-samples/tree/master/Samples/0_Introduction) (46 samples, 66 source files), [1_Utilities](https://github.com/NVIDIA/cuda-samples/tree/master/Samples/1_Utilities) (3 samples), and [5_Domain_Specific](https://github.com/NVIDIA/cuda-samples/tree/master/Samples/5_Domain_Specific). The implicit cuda_runtime.h inclusion is working on all of them and no existing passing test is broken.

In summary:
PR #2349(this PR) solves the nvcc implicitly pre-inclusion errors, even for source files that don't explicitly include them. This is the fix for the root cause you identified.

PR #2275 solves the non-self-contained headers that depend on project-specific includes (like helper_math.h, helper_cuda.h). These are provided by the specific project. The --local-headers header injection addresses this by injecting the preceding system includes from the main source file when hipifying individual local headers.

emankov:
Thank you, @ranapratap55 , for your research and explanations. So let's ship this one.

emankov (approved):
LGTM, Thanks! Please add a couple of (synthetic) tests on this feature in a separate PR.

PR #2378, opened 2026-02-10
[HIPIFY][FILE][test] Adding `cuFile` synthetic test - Part 2 - `Functions`

ranapratap55:
- Added synthetic test for `cuFile Functions`

### DIFF
--- a/tests/lit.cfg
+++ b/tests/lit.cfg
@@ -63,6 +63,7 @@ else:
 
 if not config.cuda_file_root or config.cuda_file_root == "OFF":
     config.excludes.append('cufile2hipfile.cu')
+    config.excludes.append('cufile2hipfile_functions.cu')
     print("WARN: cuFile tests are excluded because CUDA_FILE_ROOT_DIR is not specified or CUDA version < 11.4")
     warns = True
 
--- a/tests/lit.cfg
+++ b/tests/lit.cfg
@@ -61,7 +61,7 @@ else:
     if config.cutensor_version_major >= 2:
        config.excludes.append('cutensor2hiptensor_before_20000.cu')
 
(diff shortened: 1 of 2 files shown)


tests/lit.cfg
```diff
@@ -63,6 +63,7 @@ else:
 
 if not config.cuda_file_root or config.cuda_file_root == "OFF":
     config.excludes.append('cufile2hipfile.cu')
+    config.excludes.append('cufile2hipfile_functions.cu')
```
emankov, line 66: We also need to exclude this test for CUDA < 11.4

tests/unit_tests/synthetic/libraries/cufile2hipfile_functions.cu
```diff
@@ -0,0 +1,238 @@
...
+
+  // CHECK: hipFileBoolConfigParameter_t boolParam;
+  CUFileBoolConfigParameter_t boolParam;
+
+  // CHECK: hipFileStringConfigParameter_t stringParam;
+  CUFileStringConfigParameter_t stringParam;
+
+  size_t sizeTValue = 0;
+  bool boolValue = false;
+  char stringValue[256];
```
emankov: Better to make this magic number and the below once as a const variable.

emankov (changes requested):

tests/lit.cfg
ranapratap55: Added.

tests/unit_tests/synthetic/libraries/cufile2hipfile_functions.cu
ranapratap55: Updated, thanks.

emankov (approved):
LGTM, thanks!

PR #2383, opened 2026-02-11
[HIPIFY][Infra] Enable dynamic drive allocation to `Windows` build

kirthana14m:
Following changes are accumulated in this PR

1. Dynamic check of available drive on windows machine.
2. Test cluster name is reverted to linux-mi325-1gpu-ossci-rocm
3. Base ROCm CI is updated to de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
4. aomp smoke tests are added (similar to llvm-project updated)

### DIFF
--- a/.github/workflows/build_portable_linux_artifacts.yml
+++ b/.github/workflows/build_portable_linux_artifacts.yml
@@ -205,7 +205,7 @@ jobs:
           python3 build_tools/analyze_build_times.py --build-dir build
 
       - name: Configure AWS Credentials for non-forked repos
-        if: always()
+        if: ${{ github.repository == 'ROCm/TheRock' }}
         uses: aws-actions/configure-aws-credentials@61815dcd50bd041e203e49132bacad1fd04d2708 # v5.1.1
         with:
           aws-region: us-east-2
--- a/.github/workflows/build_portable_linux_artifacts.yml
+++ b/.github/workflows/build_portable_linux_artifacts.yml
@@ -204,7 +204,7 @@ jobs:
         run: |
           python3 build_tools/analyze_build_times.py --build-dir build
 
-      - name: Configure AWS Credentials for non-forked repos
+      - name: Configure AWS Credentials
         if: ${{ github.repository == 'ROCm/TheRock' }}
         uses: aws-actions/configure-aws-credentials@61815dcd50bd041e203e49132bacad1fd04d2708 # v5.1.1
         with:
--- a/.github/workflows/build_portable_linux_artifacts.yml
+++ b/.github/workflows/build_portable_linux_artifacts.yml
@@ -49,6 +49,8 @@ on:
         type: boolean
       extra_cmake_options:
         type: string
+      test_type:
+        type: string
 
 # See the details regarding permissions from the link:
 # https://github.com/aws-actions/configure-aws-credentials?tab=readme-ov-file#oidc
@@ -82,8 +84,9 @@ jobs:
         with:
           repository: "ROCm/TheRock"
           ref: 'compiler/amd-staging'
-          
+
       - name: Update Submodule Pointer to the PR
+        if: ${{ github.event_name == 'pull_request' }}
         run: |
           git config --global --add safe.directory $PWD
           # Fetch the latest commit SHA from the PR branch          
@@ -96,7 +99,6 @@ jobs:
           # Verify the pointer update
           git submodule status
           git submodule
-    
 
       - name: Install python deps
         run: |
@@ -121,10 +123,6 @@ jobs:
         run: |
           ./build_tools/health_status.py
 
-      - name: Test build_tools
-        run: |
-          python -m pytest build_tools/tests build_tools/github_actions/tests
-
       - name: Remove ununsed existing patchces
         run: |
           rm -fv patches/amd-mainline/llvm-project/0001-Ensure-to-use-libamdhip64-with-major-version.patch
@@ -134,32 +132,8 @@ jobs:
         run: |
           ./build_tools/fetch_sources.py --jobs 12
 
-      - name: "Checking out repository for llvm-project"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-        with:
-          repository: "ROCm/llvm-project"
-          path: compiler/amd-llvm
-          ref: amd-staging
-
-      - name: "Checking out repository for spirv-llvm-translator"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-        with:
-          repository: "ROCm/spirv-llvm-translator"
-          path: compiler/spirv-llvm-translator
-          ref: amd-staging
-
-      - name: Apply patches
-        run: |
-             cp -v patches/amd-mainline/llvm-project/*.patch compiler/amd-llvm
-             cd compiler/amd-llvm
-             git log -10
-             git config --global --add safe.directory $PWD
-             find . -type f -name '*.patch' -exec git apply --check {} \;
-             find . -type f -name '*.patch' -exec git apply {} \;
-             git log -15
-             cd -
-
       - name: TheRock and llvm SHA
+        if: ${{ github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
         run: |
              git config --global --add safe.directory $PWD
              git log --oneline -1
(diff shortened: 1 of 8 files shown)


emankov (changes requested):
Please see my review of #2384.

.github/workflows/build_portable_linux_artifacts.yml
```diff
@@ -121,10 +123,6 @@ jobs:
         run: |
           ./build_tools/health_status.py
 
-      - name: Test build_tools
```
emankov, line 124: In both the `Linux` and `Windows` build workflows, the `Test build_tools` was removed.

Unless these tests were moved to a separate dedicated workflow not included in this patch, you are losing `CI coverage` for them.

.github/workflows/build_windows_artifacts.yml
```diff
@@ -186,19 +190,36 @@ jobs:
...
+        env:
+          cmake_preset: ${{ inputs.build_variant_cmake_preset }}
+          amdgpu_families: ${{ inputs.amdgpu_families }}
+          package_version: ${{ inputs.package_version }}
+          extra_cmake_options: ${{ inputs.extra_cmake_options }}
+        run: |
+          # clear cache before build and after download
+          cd "${{ steps.subst.outputs.drive }}/"
+          ccache -z
+          cmake -B "B:\build" -GNinja -S "${{ steps.subst.outputs.drive }}/" --preset windows-release -DTHEROCK_AMDGPU_FAMILIES=gfx1151 "-DCMAKE_C_COMPILER=C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207/bin/Hostx64/x64/cl.exe" "-DCMAKE_CXX_COMPILER=C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207/bin/Hostx64/x64/cl.exe" "-DCMAKE_LINKER=C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207/bin/Hostx64/x64/link.exe" -DTHEROCK_BACKGROUND_BUILD_JOBS=4
```
emankov, line 217: I already saw and reviewed that in 2384: https://github.com/ROCm/HIPIFY/pull/2384#discussion_r2793343078.
Those comments were left unaddressed:

1. You should always use "" for paths instead of '', as in Windows CMD, single quotes are not string delimiters.
2. Targetting a very specific MSVS version, like 14.44.35207, is a bad solution. You should use vcvars64.bat beforehand.
3. Could you explain why and where the drive letter B has appeared?

.github/workflows/test_aomp_smoke.yml
```diff
@@ -0,0 +1,135 @@
...
+permissions:
+  contents: read
+
+jobs:
+  aomp_smoke_test:
+    name: "aomp smoke test"
+    runs-on: ${{ inputs.test_runs_on }}
+    # Running docker with cap-add and -v /lib/modiles, by recommendation of Github: https://rocm.docs.amd.com/projects/amdsmi/en/amd-staging/how-to/setup-docker-container.html
+    container:
+      image: ${{ inputs.platform == 'linux' && 'ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:4150afe4759d14822f0e3f8930e1124f26e11f68b5c7b91ec9a02b20b1ebbb98' || null }}
```
emankov, line 51: `Windows`: `null`.

Just ensure that the script `aomp/bin/run_theRockCI.sh` includes fail-safes in case it accidentally executes directly on a `Windows` runner host.

.github/workflows/build_windows_artifacts.yml
```diff
@@ -95,7 +98,35 @@ jobs:
...
+          )
+          REM L: in use — find another available drive letter
+          for %%L in (Z Y X W V U T S R Q P O N M K) do (
+            if not exist %%L:\ (
+              set "driveLetter=%%L:"
+              goto :found
+            )
+          )
+          echo No available drive letter found. So exiting
+          exit /b 1
```
emankov: [MINOR] While functionally fine, it's good practice to wrap the echo string in quotes or redirect to standard error `>&2` for better logging visibility in `GitHub Actions`.

emankov (changes requested):
Thanks for addressing some of my comments in the already closed #2384; however, there are still some issues to address in this PR.

**[IMPORTANT]** 
I'd still suggest that you revise the coupled PRs approach for `amd-staging` and `amd-mainline` for the following reasons:
1. `error-prone`
2. `common source base` vs `code duplication`
3. difficulties in reviewing them both altogether, keeping in mind different target branches

@searlmc1, @kzhuravl, jfyi.

skganesan008 (changes requested):
Please refer to the comments on https://github.com/ROCm/SPIRV-LLVM-Translator/pull/52

.github/workflows/build_portable_linux_artifacts.yml
kirthana14m: Yes, tests now live in a dedicated workflow and will be re-enabled once the testing phase is complete in a new patch.

These tests were temporarily removed to accommodate infrastructure changes.

.github/workflows/build_windows_artifacts.yml
kirthana14m: 1. Paths: Already using "" for paths on Line 217; happy to fix any remaining spots you flag.

2. MSVS : We’re using a dedicated Windows runner image that’s built and maintained by ROCKCI DevOps team. That image has a fixed, supported MSVC install (hence the 14.44.35207 path). Relying on that image is a deliberate choice for consistency and supportability in our environment. We’re not invoking a loose MSVC install on a generic host; the compiler path reflects what’s actually on the image. If we move to a different image or setup (e.g. using vcvars64.bat on a generic runner), we’ll revisit this and document it.

3. Drive letter B: B:\ is the standard build drive on our Windows runner images. For example, BUILD_DIR is set to B:\build in this workflow at line 71 so all builds use that path. It’s defined by the runner image/infrastructure, not by this repo, and keeps paths consistent across runs and easier to support.

As briefly described in the 2384 comment (Point 3)

.github/workflows/test_aomp_smoke.yml
kirthana14m: Thanks for the note. 
We’ll check and add the fail-safes to aomp/bin/run_theRockCI.sh as suggested in a follow-up change.

.github/workflows/build_windows_artifacts.yml
kirthana14m: Thanks. This has been addressed.

skganesan008:
@emankov 
Thanks for taking time to review the github action files. 

Just to give you some background, we use the CI framework from the ROCK. The backend, which is the Python framework from the ROCK is used as-is . Our goal is to not diverge from what they are doing on the backend. 

The front-end which is all these yaml files, where taken from ROCK CI and leveraged for our needs. We also let temporary workarounds and changes in to these yaml files for our needs. For example, the Windows code has a hardcoded drive (L:) as a map to source code in order to address a command line length issue seen during MIOpen compilation.  The Windows builds run on containers where we know from the infrastructure side this drive L: is not already taken up  for sure. subst command inside a container is completely invisible to the host and other containers.   The  "B:" comes from ROCI CI end and that should also be not a problem as it is running inside a container and should not have an issue as these containers are managed internally. When containers change, the change itself goes through a psdb.  At a high level, when we look at it from better software development practice, these hardcoding looks bad. 

Also, please note that  merging changes from Rock would become hard if we combine amd-staging and amd-mainline yaml files.  There is always room for improvement and abstraction.

skganesan008:
@emankov, Let's get this PR landed and  Kirthana can help follow up on improvements to this PR.

skganesan008 (approved):

.github/workflows/build_windows_artifacts.yml
emankov: > hence the 14.44.35207 path

FMPOV, it is even worse than `magic numbers`. All Environment-related variables and constants should be set in one place: the higher, the better. Otherwise, with any environment changes, cloning, whatever else, you'll need to fix those multiple paths, drive letters, and consts in multiple places.

emankov:
Hello @skganesan008,

> please note that merging changes from Rock would become hard if we combine amd-staging and amd-mainline yaml files.

Does a coupled reviewing and merging of two almost-but-not-identical changes look easier/simpler?
Or will all the upcoming changes in the dependent theROC components, including HIPIFY tools, inherit theROC issues implicitly and without review?

[Conclusion]
As long as the PR is marked as `do not merge`, you can merge it without my approval or anyone else's. I don't mind.
I reviewed it by mistake. It has to be reviewed and fixed, not in HIPIFY, but higher in theROC.

emankov:
> @emankov, Let's get this PR landed and Kirthana can help follow up on improvements to this PR.

I don't mind, but the first place to start improvements and fixes looks like theROC itself.

PR #2395, opened 2026-02-20
[HIPIFY][FILE][test][Windows] Added support for testing `cuFile` on Windows

emankov:
+ [IMP] Provide `shim headers` for OS-specific system header files
+ [IMP][Linux] Force test hipification through the fallback path in those shim header files
+ [IMP] `CUDA_FILE_ROOT_DIR` must be specified only on `Windows`, as `cuFile.h` is shipped with the `CUDA Toolkit` for `Linux` and is found as a regular CUDA file
+ [Tests] The `Header Shim` approach was tested on both `Windows` and `Linux`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -404,6 +404,8 @@ if(HIPIFY_CLANG_TESTS OR HIPIFY_CLANG_TESTS_ONLY)
 
   if(CUDA_FILE_ROOT_DIR STREQUAL "")
     set(CUDA_FILE_ROOT_DIR OFF)
+  else()
+    set(CUDA_FILE_ROOT_DIR "${CUDA_FILE_ROOT_DIR}/include")
   endif()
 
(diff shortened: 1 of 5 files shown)


derobins:
FWIW, hipFile is not supported on Windows and probably won't be for some time. AFAIK, cuFile isn't supported on Windows, either.

emankov:
> FWIW, hipFile is not supported on Windows and probably won't be for some time. AFAIK, cuFile isn't supported on Windows, either.

I know, but HIPIFY tools are `cross-platform` tools. It doesn't matter which platform the HIP files were generated on.

What is actually needed is to inform the user about this fact: 1) as a hipification `warning`, 2) in the `CUDA2HIP` documentation. It will be done in a separate PR. For this PR, it is needed to have a chance to test AMAP with the same number of sources on both OS types.

ranapratap55 (approved):
LGTM.

PR #2402, opened 2026-02-23
[HIPIFY][feature] Local header hipification via Clang PPCallbacks (experimental, opt-in) - Part 1

ranapratap55:
## Problem Statement
`hipify-clang` currently only transforms the main source file passed as input. Local (quoted) headers included in the main file remain same as CUDA code. These headers often cannot be hipified individually because they are not self-contained headers which requires types, operators, and definitions from preceding includes in the parent source.

Example: [dxtc.cu](https://github.com/NVIDIA/cuda-samples/blob/master/Samples/5_Domain_Specific/dxtc/dxtc.cu) includes `CudaMath.h`, which uses `float3` operators (`+=`, `*`, `-`) defined in `helper_math.h`. Hipifying `CudaMath.h` alone fails because these operators are unavailable. Similar cases:
- [bodysystemcpu_impl.h](https://github.com/NVIDIA/cuda-samples/blob/master/Samples/5_Domain_Specific/nbody/bodysystemcpu_impl.h)
- [MonteCarlo_reduction.cuh](https://github.com/NVIDIA/cuda-samples/blob/master/Samples/5_Domain_Specific/MonteCarloMultiGPU/MonteCarlo_reduction.cuh)
- [fastWalshTransform_kernel.cuh](https://github.com/NVIDIA/cuda-samples/blob/master/Samples/5_Domain_Specific/fastWalshTransform/fastWalshTransform_kernel.cuh)

## Solution
Two new opt-in flags enable local header hipification:
```
hipify-clang --local-headers main.cu -I ../Common/ 
hipify-clang --local-headers-recursive main.cu -I ../Common/
```

The feature is **experimental** and disabled by default. It will be promoted to default behavior in later release cycles with `--skip-local-headers` as the opt-out fallback.

Supersedes #2275 

### Clang PPCallbacks instead of regex
Include discovery is using Clang's `PPCallbacks::InclusionDirective` via preprocessing pre-pass, not regex. This correctly handles conditional compilation (`#ifdef`), macro-computed includes (`#include MACRO`), compiler search paths (`-I`, `-isystem`), and comments. 

### Fixed Compilation Database fallback
Header files which have no entry in `compile_commands.json` falls back to a `FixedCompilationDatabase` rooted at the main source file's directory so that relative paths resolve correctly.

### Cycle and duplicate detection
A `processed` set prevents re-hipification, and a `queued` set prevents duplicate work-queue entries during recursive traversal.

ToDo work (follow-up PRs):

- [ ] **Warning for non-self-contained headers**: Warn (besides compiler’s error messaging) the user about imposibility to hipify a header file, which is not self-contained.
- [ ] **Folder-level hipification**: Hipify everything in the provided folder.
- [ ] **Promote to default** (add `--skip-local-headers`)
- [ ] **Angle-bracket local header hipification** (`#include <mylib/utils.h>` via `-I`) — #2505
- [ ] **Transitive preceding includes for deeply nested recursive headers** — #2507

### DIFF
--- a/src/ArgParse.cpp
+++ b/src/ArgParse.cpp
@@ -226,13 +226,8 @@ cl::opt<bool> HipDnnSupport("hipdnn",
   cl::init(false),
   cl::cat(ToolTemplateCategory));
 
-cl::opt<bool> OptLocalHeaders("local-headers",
-  cl::desc("Enable hipification of quoted local headers (non-recursive)"),
-  cl::init(false),
-  cl::cat(ToolTemplateCategory));
-
-cl::opt<bool> OptLocalHeadersRecursive("local-headers-recursive",
-  cl::desc("Enable hipification of quoted local headers recursively"),
+cl::opt<bool> SkipLocalHeaders("skip-local-headers",
+  cl::desc("Skip implicit hipification of local (quoted) headers included in the main source file"),
   cl::init(false),
   cl::cat(ToolTemplateCategory));
 
@@ -264,8 +259,7 @@ const std::vector<std::string> hipifyOptions {
   std::string(NoWarningsUndocumented.ArgStr),
   std::string(HipifyAMAP.ArgStr),
   std::string(HipDnnSupport.ArgStr),
-  std::string(OptLocalHeaders.ArgStr),
-  std::string(OptLocalHeadersRecursive.ArgStr),
+  std::string(SkipLocalHeaders.ArgStr),
 };
 
 const std::vector<std::string> hipifyOptionsWithTwoArgs {
--- a/src/ArgParse.cpp
+++ b/src/ArgParse.cpp
@@ -226,46 +226,46 @@ cl::opt<bool> HipDnnSupport("hipdnn",
   cl::init(false),
   cl::cat(ToolTemplateCategory));
 
-cl::opt<bool> OptLocalHeaders(
-    "local-headers",
-    cl::desc("Enable hipification of quoted local headers (non-recursive)"),
-    cl::init(false), cl::cat(ToolTemplateCategory));
+cl::opt<bool> OptLocalHeaders("local-headers",
+  cl::desc("Enable hipification of quoted local headers (non-recursive)"),
+  cl::init(false),
+  cl::cat(ToolTemplateCategory));
 
-cl::opt<bool> OptLocalHeadersRecursive(
-    "local-headers-recursive",
-    cl::desc("Enable hipification of quoted local headers recursively"),
-    cl::init(false), cl::cat(ToolTemplateCategory));
+cl::opt<bool> OptLocalHeadersRecursive("local-headers-recursive",
+  cl::desc("Enable hipification of quoted local headers recursively"),
+  cl::init(false),
+  cl::cat(ToolTemplateCategory));
 
 cl::extrahelp CommonHelp(ct::CommonOptionsParser::HelpMessage);
 
-const std::vector<std::string> hipifyOptions{
-    std::string(PrintStatsCSV.ArgStr),
-    std::string(PrintStats.ArgStr),
-    std::string(SkipExcludedPPConditionalBlocks.ArgStr),
-    std::string(DefaultPreprocessor.ArgStr),
-    std::string(HipKernelExecutionSyntax.ArgStr),
-    std::string(CudaKernelExecutionSyntax.ArgStr),
-    std::string(GeneratePerl.ArgStr),
-    std::string(GeneratePython.ArgStr),
-    std::string(TranslateToRoc.ArgStr),
-    std::string(TranslateToMIOpen.ArgStr),
-    std::string(GenerateMarkdown.ArgStr),
-    std::string(GenerateCSV.ArgStr),
-    std::string(NoBackup.ArgStr),
-    std::string(NoOutput.ArgStr),
-    std::string(Inplace.ArgStr),
-    std::string(Examine.ArgStr),
-    std::string(SaveTemps.ArgStr),
-    std::string(DocFormat.ArgStr),
-    std::string(DocRoc.ArgStr),
-    std::string(Experimental.ArgStr),
-    std::string(Versions.ArgStr),
-    std::string(NoUndocumented.ArgStr),
-    std::string(NoWarningsUndocumented.ArgStr),
-    std::string(HipifyAMAP.ArgStr),
-    std::string(HipDnnSupport.ArgStr),
-    std::string(OptLocalHeaders.ArgStr),
-    std::string(OptLocalHeadersRecursive.ArgStr),
+const std::vector<std::string> hipifyOptions {
+  std::string(PrintStatsCSV.ArgStr),
+  std::string(PrintStats.ArgStr),
+  std::string(SkipExcludedPPConditionalBlocks.ArgStr),
+  std::string(DefaultPreprocessor.ArgStr),
+  std::string(HipKernelExecutionSyntax.ArgStr),
+  std::string(CudaKernelExecutionSyntax.ArgStr),
+  std::string(GeneratePerl.ArgStr),
+  std::string(GeneratePython.ArgStr),
+  std::string(TranslateToRoc.ArgStr),
+  std::string(TranslateToMIOpen.ArgStr),
+  std::string(GenerateMarkdown.ArgStr),
+  std::string(GenerateCSV.ArgStr),
+  std::string(NoBackup.ArgStr),
+  std::string(NoOutput.ArgStr),
+  std::string(Inplace.ArgStr),
+  std::string(Examine.ArgStr),
+  std::string(SaveTemps.ArgStr),
+  std::string(DocFormat.ArgStr),
+  std::string(DocRoc.ArgStr),
+  std::string(Experimental.ArgStr),
+  std::string(Versions.ArgStr),
+  std::string(NoUndocumented.ArgStr),
+  std::string(NoWarningsUndocumented.ArgStr),
+  std::string(HipifyAMAP.ArgStr),
+  std::string(HipDnnSupport.ArgStr),
+  std::string(OptLocalHeaders.ArgStr),
+  std::string(OptLocalHeadersRecursive.ArgStr),
 };
 
 const std::vector<std::string> hipifyOptionsWithTwoArgs {
--- a/src/ArgParse.cpp
+++ b/src/ArgParse.cpp
@@ -226,40 +226,46 @@ cl::opt<bool> HipDnnSupport("hipdnn",
   cl::init(false),
   cl::cat(ToolTemplateCategory));
 
-cl::opt<bool> SkipLocalHeaders("skip-local-headers",
-  cl::desc("Skip implicit hipification of local (quoted) headers included in the main source file"),
-  cl::init(false),
-  cl::cat(ToolTemplateCategory));
+cl::opt<bool> OptLocalHeaders(
+    "local-headers",
+    cl::desc("Enable hipification of quoted local headers (non-recursive)"),
+    cl::init(false), cl::cat(ToolTemplateCategory));
+
+cl::opt<bool> OptLocalHeadersRecursive(
+    "local-headers-recursive",
+    cl::desc("Enable hipification of quoted local headers recursively"),
+    cl::init(false), cl::cat(ToolTemplateCategory));
 
 cl::extrahelp CommonHelp(ct::CommonOptionsParser::HelpMessage);
 
-const std::vector<std::string> hipifyOptions {
-  std::string(PrintStatsCSV.ArgStr),
-  std::string(PrintStats.ArgStr),
-  std::string(SkipExcludedPPConditionalBlocks.ArgStr),
-  std::string(DefaultPreprocessor.ArgStr),
-  std::string(HipKernelExecutionSyntax.ArgStr),
-  std::string(CudaKernelExecutionSyntax.ArgStr),
-  std::string(GeneratePerl.ArgStr),
-  std::string(GeneratePython.ArgStr),
-  std::string(TranslateToRoc.ArgStr),
-  std::string(TranslateToMIOpen.ArgStr),
-  std::string(GenerateMarkdown.ArgStr),
-  std::string(GenerateCSV.ArgStr),
-  std::string(NoBackup.ArgStr),
-  std::string(NoOutput.ArgStr),
-  std::string(Inplace.ArgStr),
-  std::string(Examine.ArgStr),
-  std::string(SaveTemps.ArgStr),
-  std::string(DocFormat.ArgStr),
-  std::string(DocRoc.ArgStr),
-  std::string(Experimental.ArgStr),
-  std::string(Versions.ArgStr),
-  std::string(NoUndocumented.ArgStr),
-  std::string(NoWarningsUndocumented.ArgStr),
-  std::string(HipifyAMAP.ArgStr),
-  std::string(HipDnnSupport.ArgStr),
-  std::string(SkipLocalHeaders.ArgStr),
+const std::vector<std::string> hipifyOptions{
+    std::string(PrintStatsCSV.ArgStr),
+    std::string(PrintStats.ArgStr),
+    std::string(SkipExcludedPPConditionalBlocks.ArgStr),
+    std::string(DefaultPreprocessor.ArgStr),
+    std::string(HipKernelExecutionSyntax.ArgStr),
+    std::string(CudaKernelExecutionSyntax.ArgStr),
+    std::string(GeneratePerl.ArgStr),
+    std::string(GeneratePython.ArgStr),
+    std::string(TranslateToRoc.ArgStr),
+    std::string(TranslateToMIOpen.ArgStr),
+    std::string(GenerateMarkdown.ArgStr),
+    std::string(GenerateCSV.ArgStr),
+    std::string(NoBackup.ArgStr),
+    std::string(NoOutput.ArgStr),
+    std::string(Inplace.ArgStr),
+    std::string(Examine.ArgStr),
+    std::string(SaveTemps.ArgStr),
+    std::string(DocFormat.ArgStr),
+    std::string(DocRoc.ArgStr),
+    std::string(Experimental.ArgStr),
+    std::string(Versions.ArgStr),
+    std::string(NoUndocumented.ArgStr),
+    std::string(NoWarningsUndocumented.ArgStr),
+    std::string(HipifyAMAP.ArgStr),
+    std::string(HipDnnSupport.ArgStr),
+    std::string(OptLocalHeaders.ArgStr),
+    std::string(OptLocalHeadersRecursive.ArgStr),
 };
 
 const std::vector<std::string> hipifyOptionsWithTwoArgs {
--- a/src/LocalHeader.cpp
+++ b/src/LocalHeader.cpp
@@ -186,13 +186,16 @@ bool hipifyLocalHeaders(const std::string &mainSourceAbsPath,
            << ": " << sys::path::filename(initial[i]) << "\n";
   }
 
-  std::vector<std::string> work(initial.begin(), initial.end());
+  std::vector<std::pair<std::string, std::string>> work;
+  for (const auto &h : initial) {
+    work.push_back({h, mainSourceAbsPath});
+  }
   std::set<std::string> processed;
   size_t total = initial.size();
   size_t current = 0;
 
   while (!work.empty()) {
-    std::string hdr = work.back();
+    auto [hdr, parentPath] = work.back();
     work.pop_back();
     if (processed.count(hdr)) {
       outs() << sHipify << sWarning
@@ -212,13 +215,13 @@ bool hipifyLocalHeaders(const std::string &mainSourceAbsPath,
 
     std::string hipOut = hdr + ".hip";
     std::vector<std::string> precedingIncludes;
-    collectPrecedingIncludes(mainSourceAbsPath, hdr, precedingIncludes);
+    collectPrecedingIncludes(parentPath, hdr, precedingIncludes);
 
     outs() << "\n" << sHipify << "Hipifying local header [" << current
            << "/" << total << "]: " << sys::path::filename(hdr) << "\n";
 
     bool ok = hipifySingleSource(hdr, hipOut, compDB, OptionsParserPtr,
-                                  hipify_exe, mainSourceAbsPath, false,
+                                  hipify_exe, parentPath, false,
                                   precedingIncludes);
 
     if (!ok) {
@@ -240,7 +243,7 @@ bool hipifyLocalHeaders(const std::string &mainSourceAbsPath,
           std::string abs;
           if (resolveLocalIncludeInternal(hdr, rel, abs) &&
               !processed.count(abs)) {
-            work.push_back(abs);
+            work.push_back({abs, hdr});
             newHeaders.push_back(abs);
           }
         }
--- a/src/LocalHeader.cpp
+++ b/src/LocalHeader.cpp
@@ -1,3 +1,25 @@
+/*
+Copyright (c) 2026 - present Advanced Micro Devices, Inc. All rights reserved.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
 #include "LocalHeader.h"
 #include "LLVMCompat.h"
 
--- a/src/LocalHeader.cpp
+++ b/src/LocalHeader.cpp
@@ -1,4 +1,5 @@
 #include "LocalHeader.h"
+#include "LLVMCompat.h"
 
 #include <sstream>
 #include <regex>
@@ -10,8 +11,6 @@
 #include "llvm/Support/Path.h"
 #include "llvm/Support/MemoryBuffer.h"
 #include "llvm/Support/raw_ostream.h"
-
-#include "LLVMCompat.h"
 
 using namespace clang;
 using namespace clang::tooling;
@@ -46,6 +45,9 @@ namespace {
   static const std::regex LocalIncludeRe(
       R"(^\s*#\s*include\s*\"([^\"\n]+)\"\s*(?://.*)?$)", std::regex::ECMAScript);
 
+  static const std::regex SystemIncludeRe(
+      R"(^\s*#\s*include\s*<([^>\n]+)>)", std::regex::ECMAScript);
+
   bool readFile(const std::string &path, std::string &out) {
     auto MBOrErr = llvm::MemoryBuffer::getFile(path);
     if (!MBOrErr) return false;
@@ -67,7 +69,41 @@ namespace {
     }
     return false;
   }
-} 
+
+  bool collectPrecedingIncludes(const std::string &mainSourceAbspath,
+                                const std::string &targetHeaderAbspath,
+                                std::vector<std::string> &outIncludes) {
+    std::string mainSourceContent;
+    if (!readFile(mainSourceAbspath, mainSourceContent)) {
+      errs() << sHipify << sError << "Cannot read source files: "
+            << mainSourceAbspath << "\n";
+      return false;
+    }
+
+    std::string targetFileName = std::string(sys::path::filename(targetHeaderAbspath));
+    std::istringstream iss(mainSourceContent);
+    std::string line;
+    std::smatch m;
+
+    while (std::getline(iss, line)) {
+      if (std::regex_match(line, m, LocalIncludeRe)) {
+        std::string quotedName = m[1].str();
+        std::string quotedFileName = std::string(sys::path::filename(quotedName));
+        if (quotedFileName == targetFileName)
+          break;
+        std::string absPath;
+        if (resolveLocalIncludeInternal(mainSourceAbspath, quotedName, absPath))
+          outIncludes.push_back(absPath);
+      }
+
+      if (std::regex_search(line, m, SystemIncludeRe)) {
+        outIncludes.push_back(m[1].str());
+      }
+    }
+
+    return true;
+  }
+}
 
 bool resolveLocalInclude(const std::string &mainSourceAbsPath,
                          const std::string &includeToken,
@@ -116,52 +152,87 @@ bool hipifyLocalHeaders(const std::string &mainSourceAbsPath,
   }
   
   if (initial.empty()) {
-    outs() << sHipify << "No local headers detected in " << mainSourceAbsPath << "\n";
+    outs() << "\n" << sHipify << "No local headers detected in "
+           << sys::path::filename(mainSourceAbsPath) << "\n";
     return true;
+  }
+
+  outs() << "\n" << sHipify << "Local headers found: " << initial.size()
+         << " in " << sys::path::filename(mainSourceAbsPath) << "\n";
+  for (size_t i = 0; i < initial.size(); ++i) {
+    outs() << (i + 1) << "/" << initial.size()
+           << ": " << sys::path::filename(initial[i]) << "\n";
   }
 
   std::vector<std::string> work(initial.begin(), initial.end());
   std::set<std::string> processed;
+  size_t total = initial.size();
+  size_t current = 0;
 
   while (!work.empty()) {
     std::string hdr = work.back();
     work.pop_back();
     if (processed.count(hdr)) {
-      errs() << sHipify << sWarning << "Duplicate local header reference ignored: " << hdr << "\n";
+      outs() << sHipify << sWarning
+             << "Duplicate local header reference ignored: "
+             << sys::path::filename(hdr) << "\n";
       continue;
     }
     processed.insert(hdr);
+    ++current;
 
     std::string original;
     if (!readFile(hdr, original)) {
-      errs() << sHipify << sError << "Cannot read header: " << hdr << "\n";
+      errs() << "\n" << sHipify << sError
+             << "Cannot read header: " << sys::path::filename(hdr) << "\n";
       continue;
     }
 
     std::string hipOut = hdr + ".hip";
+    std::vector<std::string> precedingIncludes;
+    collectPrecedingIncludes(mainSourceAbsPath, hdr, precedingIncludes);
+
+    outs() << "\n" << sHipify << "Hipifying local header [" << current
+           << "/" << total << "]: " << sys::path::filename(hdr) << "\n";
+
     bool ok = hipifySingleSource(hdr, hipOut, compDB, OptionsParserPtr,
-                                  hipify_exe, mainSourceAbsPath, false);
+                                  hipify_exe, mainSourceAbsPath, false,
+                                  precedingIncludes);
 
     if (!ok) {
-      errs() << sHipify << sError << "Hipify failed for header: " << hdr << "\n";
+      errs() << "\n" << sHipify << sError
+             << "Hipify failed for header [" << current << "/" << total
+             << "]: " << sys::path::filename(hdr) << "\n";
       return false;
     }
+    outs() << sHipify << "Successfully hipified header file" << "\n";
 
     if (recursive) {
       std::smatch m;
       std::istringstream iss(original);
       std::string line;
+      std::vector<std::string> newHeaders;
       while (std::getline(iss, line)) {
         if (std::regex_match(line, m, LocalIncludeRe)) {
           std::string rel = m[1].str();
           std::string abs;
           if (resolveLocalIncludeInternal(hdr, rel, abs) &&
-              !processed.count(abs))
+              !processed.count(abs)) {
             work.push_back(abs);
+            newHeaders.push_back(abs);
+          }
         }
+      }
+      if (!newHeaders.empty()) {
+        total += newHeaders.size();
+        outs() << sHipify << "  Recursive: found " << newHeaders.size()
+               << " additional local header(s) in "
+               << sys::path::filename(hdr) << "\n";
       }
     }
   }
 
+  outs() << "\n" << sHipify << "Local header hipification complete: "
+         << processed.size() << " header(s) processed.\n";
   return true;
 }
--- a/src/LocalHeader.cpp
+++ b/src/LocalHeader.cpp
@@ -24,15 +24,17 @@ THE SOFTWARE.
 #include "LLVMCompat.h"
 
 #include <memory>
-#include <set>
+#include <string>
 #include <vector>
 
+#include "clang/Basic/SourceManager.h"
 #include "clang/Frontend/CompilerInstance.h"
 #include "clang/Frontend/FrontendActions.h"
 #include "clang/Lex/PPCallbacks.h"
 #include "clang/Lex/Preprocessor.h"
 #include "clang/Tooling/CompilationDatabase.h"
 #include "clang/Tooling/Tooling.h"
+#include "llvm/ADT/StringSet.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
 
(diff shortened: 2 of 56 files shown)


emankov:
Can we say that the main purpose of this PR is to allow hipifying header files individually as main translation units?

ranapratap55:
> Can we say that the main purpose of this PR is to allow hipifying header files individually as main translation units?

Yes, correct. The main purpose is to allow hipifying local header files as main translation units. We are doing it by injecting their preceding includes from the parent source file using Clang's `-include ArgumentsAdjusters` which is to provide the necessary compilation context. 

This feature is enabled via `--local-headers / --local-headers-recursive` and only applies to local (quoted) headers referenced in the main source file.

emankov:
> the parent source

How do we guarantee finding the parent source(s) for a particular header file so we can hipify it as the main translation unit?

ranapratap55:
> > the parent source
> 
> How do we guarantee finding the parent source(s) for a particular header file so we can hipify it as the main translation unit?

We don't need to discover/guess the parent source file. The user provides it explicitly while hipifying. The main source file passed to hipify-clang is always the parent. `--local-headers` simply extends hipification to that source file's local headers, using the source file itself as the context for preceding includes.

For example: `hipify-clang --local-headers sample.cu`: here `sample.cu` is known parent for its local headers (`header1.h, header2.h`, etc.,).

emankov:
> hipify-clang --local-headers sample.cu

That is what worries me most.

1. The user should know or learn about the main source file, or, in general, the source files that include that particular header file.
2. sample.cu is not a header file. Here we have misleading information for the end users.
3. We need to explain somewhere, somehow, this tricky feature (actually a workaround).
4. Finally, it makes hipify-clang even more complicated in use, which we definitely do not want, as users already consider hipify-clang complicated enough to prefer hipify-perl usage instead. And we want the opposite. We want users to prefer hipify-clang. It means that we need to simplify it AMAP and do more work under the hood implicitly.

Technically, the main problem is that we are trying to treat a file, a header file, which **is not self-contained**, as a source file for hipification (actually, compilation).

I want us to think about other approaches here.
For instance:

1. Warn (besides compiler’s error messaging) the user about imposibility to hipify a header file, which is not self-contained.
2. Implement the ability to implicitly hipify all header files that are included in the main file.
3. Hipify everything in the provided folder.

ranapratap55:
> > hipify-clang --local-headers sample.cu
> 
> That is what worries me most.
> 
> 1. The user should know or learn about the main source file, or, in general, the source files that include that particular header file.
> 2. sample.cu is not a header file. Here we have misleading information for the end users.
> 3. We need to explain somewhere, somehow, this tricky feature (actually a workaround).
> 4. Finally, it makes hipify-clang even more complicated in use, which we definitely do not want, as users already consider hipify-clang complicated enough to prefer hipify-perl usage instead. And we want the opposite. We want users to prefer hipify-clang. It means that we need to simplify it AMAP and do more work under the hood implicitly.
> 
> Technically, the main problem is that we are trying to treat a file, a header file, which **is not self-contained**, as a source file for hipification (actually, compilation).
> 
> I want us to think about other approaches here. For instance:
> 
> 1. Warn (besides compiler’s error messaging) the user about imposibility to hipify a header file, which is not self-contained.
> 2. Implement the ability to implicitly hipify all header files that are included in the main file.
> 3. Hipify everything in the provided folder.

As per our discussions/suggestions, I've reworked this PR to implement suggestion 2.
Local header hipification is now implicit by default. `hipify-clang sample.cu` automatically detects and hipifies all local headers recursively, no flags needed. Added `--skip-local-headers` as the single opt-out mechanism. 

I will raise 2 separate PRs as a follow up for:
1.  Warn the user for non-self-contained headers
2.  Folder-level hipification

Updated PR title and description to reflect the changes.

src/LocalHeader.cpp
```diff
@@ -46,6 +67,9 @@ namespace {
   static const std::regex LocalIncludeRe(
       R"(^\s*#\s*include\s*\"([^\"\n]+)\"\s*(?://.*)?$)", std::regex::ECMAScript);
```
emankov: Missed it previously.

This regex fails if an include has a C-style block comment like `#include "a.h" /* comment */` due to the `$` anchor. Just drop the `$` anchor (like in `SystemIncludeRe`).

src/LocalHeader.cpp
```diff
@@ -116,52 +174,90 @@ bool hipifyLocalHeaders(const std::string &mainSourceAbsPath,
...
     work.pop_back();
     if (processed.count(hdr)) {
-      errs() << sHipify << sWarning << "Duplicate local header reference ignored: " << hdr << "\n";
+      outs() << sHipify << sWarning
+             << "Duplicate local header reference ignored: "
+             << sys::path::filename(hdr) << "\n";
       continue;
     }
     processed.insert(hdr);
+    ++current;
```
emankov: [Issue]
-  If a common header (e.g., `types.h`) is included by both `A.h` and `B.h`, it will be pushed to work twice. 
- The total is incorrectly incremented twice. 
- When the duplicate is eventually popped, it hits the `processed.count(hdr)` check and continues. 
- This results in skewed terminal progress output (e.g., [4/5] complete where 5 is never reached), and it spams the console with `Duplicate local header reference ignored` warnings for valid C++ inclusion patterns.

src/ArgParse.cpp
```diff
@@ -226,13 +226,8 @@ cl::opt<bool> HipDnnSupport("hipdnn",
...
 
-cl::opt<bool> OptLocalHeaders("local-headers",
-  cl::desc("Enable hipification of quoted local headers (non-recursive)"),
-  cl::init(false),
-  cl::cat(ToolTemplateCategory));
-
-cl::opt<bool> OptLocalHeadersRecursive("local-headers-recursive",
-  cl::desc("Enable hipification of quoted local headers recursively"),
+cl::opt<bool> SkipLocalHeaders("skip-local-headers",
+  cl::desc("Skip implicit hipification of local (quoted) headers included in the main source file"),
```
emankov: What if the "local" headers are included with `<>`?

src/LocalHeader.cpp
```diff
@@ -116,52 +174,90 @@ bool hipifyLocalHeaders(const std::string &mainSourceAbsPath,
...
+    std::vector<std::string> precedingIncludes;
+    collectPrecedingIncludes(parentPath, hdr, precedingIncludes);
+
+    outs() << "\n" << sHipify << "Hipifying local header [" << current
+           << "/" << total << "]: " << sys::path::filename(hdr) << "\n";
+
     bool ok = hipifySingleSource(hdr, hipOut, compDB, OptionsParserPtr,
-                                  hipify_exe, mainSourceAbsPath, false);
+                                  hipify_exe, parentPath, false,
+                                  precedingIncludes);
```
emankov: Don't quite understand why the `parentPath` is used instead of the `mainSourceAbsPath`.
Anyway, working with the `compile_commands.json` database breaks here.

src/LocalHeader.cpp
```diff
@@ -67,7 +91,41 @@ namespace {
...
+      }
+
+      if (std::regex_search(line, m, SystemIncludeRe)) {
+        outIncludes.push_back(m[1].str());
+      }
+    }
+
+    return true;
+  }
+}
```
emankov: Looks like the `collectPrecedingIncludes` only scans the immediate `parentPath`.
What about deeply nested headers?

src/LocalHeader.cpp
```diff
@@ -67,7 +91,41 @@ namespace {
...
+    std::istringstream iss(mainSourceContent);
+    std::string line;
+    std::smatch m;
+
+    while (std::getline(iss, line)) {
+      if (std::regex_match(line, m, LocalIncludeRe)) {
+        std::string quotedName = m[1].str();
+        std::string quotedFileName = std::string(sys::path::filename(quotedName));
+        if (quotedFileName == targetFileName)
+          break;
```
emankov: Isn't it a premature Loop Break on filename collision?

emankov (changes requested):

src/LocalHeader.cpp
ranapratap55: Fixed in latest revision.

src/LocalHeader.cpp
ranapratap55: Fixed in latest. I've added a `queued` set which prevents the push from happening at all. I've also fixed the warning.

src/LocalHeader.cpp
ranapratap55: Yes, thanks for this. Fixed by comparing full absolute paths (`absPath == targetHeaderAbspath`) instead.

src/ArgParse.cpp
ranapratap55: I believe this is out of scope. Angled-bracket includes files are resolved using `-I` search paths. They represent system headers or external library dependencies. 

But I do think there is edge case like, if a project uses `#include <mylib/utils.h>` via `-I project/src` flag, that would require resolving each angled-bracket against every `-I` path. That is a meaningful work and requires a separate PR of its own.

src/LocalHeader.cpp
ranapratap55: >Don't quite understand why the parentPath is used instead of the mainSourceAbsPath.

When headers are discovered recursively, using `mainSourceAbsPath` is factually wrong. Take below:
```
main.cu --> sample.h --> helper.h

(if helper.h depends on types.h, which is included in sample.h)
```
When hipifying `helper.h` we need includes that appear before it in `sample.h`, which is `types.h`. If we call `parentPath`, we get `main.cu` and finds no mention of `helper.h`, then `helper.h` fails to compile because `types.h` was never injected.

ranapratap55: >Anyway, working with the compile_commands.json database breaks here.

Thanks for this. I've added a helper function that selects right context. The compile_commands lookup now falls back to `mainSourceAbsPath` when `parentPath` has no database entry.

src/LocalHeader.cpp
ranapratap55: yes, correct. 
`collectPrecedingIncludes` scans only one level (the immediate parent). This is intentional for this PR, it was designed for (`main.cu → A.h → B.h`) where `B.h` depends on explicitly included before it in `A.h` and `collectPrecedingIncludes(A.h, B.h)` finds exactly what `B.h` needs.

I considered collecting full ancestor chain recursively (for nested headers) but that results in over-injection. When two of those injected headers define same symbol, we get redefinition errors.

I would like to take this (deeply nested headers) as a follow-up PR. We track the full ancestors list and modify from `{hdr, parentPath}` to `{hdr, vector<parentChain>}`. And only injects that appear in ancestors(parents) up to the nearest file that explicitly includes the current target. This avoids over-injection issue.

ranapratap55:
Follow up work
- #2507: Transitive preceding includes for deeply nested recursive headers
- #2505: Angle-bracket local header hipification (`#include <>`)

src/ArgParse.cpp
ranapratap55: #2505: Angle-bracket local header hipification (#include <>)

src/LocalHeader.cpp
ranapratap55: #2507: Transitive preceding includes for deeply nested recursive headers

src/LocalHeader.cpp
```diff
@@ -116,52 +186,97 @@ bool hipifyLocalHeaders(const std::string &mainSourceAbsPath,
...
+  size_t total = initial.size();
+  size_t current = 0;
+
+  for (const auto &h: initial) {
+    queued.insert(h);
+  }
 
   while (!work.empty()) {
-    std::string hdr = work.back();
+    auto [hdr, parentPath] = work.back();
```
emankov: `Structured Bindings` (`C++17`) would be a great solution here if we had not needed to support different LLVM versions. For instance, `LLVM < 16` uses `C++14` by default, so this will cause HIPIFY itself to fail to build.

src/LocalHeader.cpp
```diff
@@ -44,7 +65,10 @@ static bool pathExists(const std::string &p) {
 
 namespace {
   static const std::regex LocalIncludeRe(
-      R"(^\s*#\s*include\s*\"([^\"\n]+)\"\s*(?://.*)?$)", std::regex::ECMAScript);
+      R"(^\s*#\s*include\s*\"([^\"\n]+)\"\s*(?://.*)?)", std::regex::ECMAScript);
+
+  static const std::regex SystemIncludeRe(
+      R"(^\s*#\s*include\s*<([^>\n]+)>)", std::regex::ECMAScript);
```
emankov: I don't quite understand the removal of an anchor `$`.
In `LocalIncludeRe`, please consider using:
```regex
R"(^\s*#\s*include\s*"([^"\n]+)")"
```
similarly to `regex`, already used in `SystemIncludeRe`:
```regex
R"(^\s*#\s*include\s*<([^>\n]+)>)"
```

emankov (changes requested):

src/LocalHeader.cpp
ranapratap55: Thanks for this. I've updated in latest revision.

src/LocalHeader.cpp
ranapratap55: I've updated in latest revision.

emankov:
The added tests cover basic recursion and depth-1 injection of “includes”. However, the test that verifies the transitive context is preserved during recursive traversal is still needed.

Please add a test that includes a non-self-contained header with a recursion depth of 2 or more. For example, the `root .cu` file includes `<cmath>` and `parent.h`; `parent.h` includes `child.h`; and `child.h` uses `sqrtf` but does not include `<cmath>` itself. This is essential to ensure that the inherited compilation context is correctly propagated down the inclusion chain and is not dropped when processing nested files.

emankov:
I honestly believe we need to preserve `-local-headers` and `--local-headers-recursive`, since the `feature` you're introducing is experimental and shouldn't be used by default. What do you think, @ranapratap55?

ranapratap55:
> The added tests cover basic recursion and depth-1 injection of “includes”. However, the test that verifies the transitive context is preserved during recursive traversal is still needed.
> 
> Please add a test that includes a non-self-contained header with a recursion depth of 2 or more. For example, the `root .cu` file includes `<cmath>` and `parent.h`; `parent.h` includes `child.h`; and `child.h` uses `sqrtf` but does not include `<cmath>` itself. This is essential to ensure that the inherited compilation context is correctly propagated down the inclusion chain and is not dropped when processing nested files.

I've adapted to use `<algorithm>` + `std::sort`

- `transitive_system_include.cu`: includes `<algorithm>` and` transitive_parent.h`
- `transitive_parent.h`: includes `transitive_child.h` and `<cuda_runtime.h>`, but no `<algorithm>`
- `transitive_child.h`: uses` std::sort` and `cudaMalloc`, includes neither `<algorithm>` nor `<cuda_runtime.h>`

`collectAncestorSystemIncludes` walks the ancestor chain and collects `<algorithm>` from the main src file, then injects it before compiling `transitive_child.h`. If this propagation breaks, the test fails directly.

ranapratap55:
> I honestly believe we need to preserve `-local-headers` and `--local-headers-recursive`, since the `feature` you're introducing is experimental and shouldn't be used by default. What do you think, @ranapratap55?

Thanks for raising this and it is a fair concern. 

Based on our earlier discussions, we collectively moved toward implicit-by-default to lower the barrier for users. The current revision already has multiple layers in place for end users like 

- **Opt-out flag**: `--skip-local-headers` for users who don't want automatic hipification
- **Non-self-contained header warning**: `[HIPIFY] warning: non-self-contained header` is emitted when a header cannot be hipified in isolation and with a note warning the user to run the parent source file instead

And I am thinking to add below 

- [ ] **Auto-hipification notice**: an info-level message `[HIPIFY] note: local headers detected: hipifying automatically. Use --skip-local-headers to disable`.


With these measures together I believe the implicit behavior is transparent. That said, I am happy to address any specific scenario where this causes harm directly rather than revert the design.

What do you think? @emankov

emankov:
> > I honestly believe we need to preserve `-local-headers` and `--local-headers-recursive`, since the `feature` you're introducing is experimental and shouldn't be used by default. What do you think, @ranapratap55?
> 
> Thanks for raising this and it is a fair concern.
> 
> Based on our earlier discussions, we collectively moved toward implicit-by-default to lower the barrier for users. The current revision already has multiple layers in place for end users like
> 
> * **Opt-out flag**: `--skip-local-headers` for users who don't want automatic hipification
> * **Non-self-contained header warning**: `[HIPIFY] warning: non-self-contained header` is emitted when a header cannot be hipified in isolation and with a note warning the user to run the parent source file instead
> 
> And I am thinking to add below
> 
> * [ ]  **Auto-hipification notice**: an info-level message `[HIPIFY] note: local headers detected: hipifying automatically. Use --skip-local-headers to disable`.
> 
> With these measures together I believe the implicit behavior is transparent. That said, I am happy to address any specific scenario where this causes harm directly rather than revert the design.
> 
> What do you think? @emankov



> > I honestly believe we need to preserve `-local-headers` and `--local-headers-recursive`, since the `feature` you're introducing is experimental and shouldn't be used by default. What do you think, @ranapratap55?
> 
> Thanks for raising this and it is a fair concern.
> 
> Based on our earlier discussions, we collectively moved toward implicit-by-default to lower the barrier for users. The current revision already has multiple layers in place for end users like
> 
> * **Opt-out flag**: `--skip-local-headers` for users who don't want automatic hipification
> * **Non-self-contained header warning**: `[HIPIFY] warning: non-self-contained header` is emitted when a header cannot be hipified in isolation and with a note warning the user to run the parent source file instead
> 
> And I am thinking to add below
> 
> * [ ]  **Auto-hipification notice**: an info-level message `[HIPIFY] note: local headers detected: hipifying automatically. Use --skip-local-headers to disable`.
> 
> With these measures together I believe the implicit behavior is transparent. That said, I am happy to address any specific scenario where this causes harm directly rather than revert the design.
> 
> What do you think? @emankov



> > I honestly believe we need to preserve `-local-headers` and `--local-headers-recursive`, since the `feature` you're introducing is experimental and shouldn't be used by default. What do you think, @ranapratap55?
> 
> Thanks for raising this and it is a fair concern.
> 
> Based on our earlier discussions, we collectively moved toward implicit-by-default to lower the barrier for users. The current revision already has multiple layers in place for end users like
> 
> * **Opt-out flag**: `--skip-local-headers` for users who don't want automatic hipification
> * **Non-self-contained header warning**: `[HIPIFY] warning: non-self-contained header` is emitted when a header cannot be hipified in isolation and with a note warning the user to run the parent source file instead
> 
> And I am thinking to add below
> 
> * [ ]  **Auto-hipification notice**: an info-level message `[HIPIFY] note: local headers detected: hipifying automatically. Use --skip-local-headers to disable`.
> 
> With these measures together I believe the implicit behavior is transparent. That said, I am happy to address any specific scenario where this causes harm directly rather than revert the design.
> 
> What do you think? @emankov



> > I honestly believe we need to preserve `-local-headers` and `--local-headers-recursive`, since the `feature` you're introducing is experimental and shouldn't be used by default. What do you think, @ranapratap55?
> 
> Thanks for raising this and it is a fair concern.
> 
> Based on our earlier discussions, we collectively moved toward implicit-by-default to lower the barrier for users. The current revision already has multiple layers in place for end users like
> 
> * **Opt-out flag**: `--skip-local-headers` for users who don't want automatic hipification
> * **Non-self-contained header warning**: `[HIPIFY] warning: non-self-contained header` is emitted when a header cannot be hipified in isolation and with a note warning the user to run the parent source file instead
> 
> And I am thinking to add below
> 
> * [ ]  **Auto-hipification notice**: an info-level message `[HIPIFY] note: local headers detected: hipifying automatically. Use --skip-local-headers to disable`.
> 
> With these measures together I believe the implicit behavior is transparent. That said, I am happy to address any specific scenario where this causes harm directly rather than revert the design.
> 
> What do you think? @emankov

I still believe that experimental features should stay experimental for at least one release. Meanwhile, it might be thoroughly tested. Next step: inform end users in the documentation that, with the next release, this particular feature will be enabled by default, and for fallback functionality, here are the options to set explicitly.

src/LocalHeader.cpp
```diff
@@ -67,7 +91,77 @@ namespace {
...
+        }
+      }
+      if (std::regex_search(line, m, SystemIncludeRe)) {
+        std::string sysInclude = m[1].str();
+        if (seen.insert(sysInclude).second)
+          outIncludes.push_back(sysInclude);
+      }
+    }
+    return true;
+  }
```
emankov: **[Issue]**
Clang evaluates compiler flags strictly from left to right. As a result, Clang will process the Parent's headers before the Root's headers. This might completely break C++ dependencies.
**[Solution]**
The algorithm must build the preceding Includes array in chronological order, top-down (from the oldest ancestor to the immediate parent).

emankov:
**[IMPORTANT]**

After reviewing this regex-based solution for header tree calculation, I decided it is not a good solution. Using regex in hipify-perl is a forced decision, but **using a regex-based approach with a C++ compiler under the hood can't be treated as a good solution.** 

**[Reasons]**
+ Parsing C/C++ includes with regular expressions is fundamentally brittle and serves as a temporary workaround.
+ Regex has major blind spots that cannot be solved without rewriting a full C++ preprocessor:
   - **Conditional Compilation**: Regex blindly captures `#include` directives inside `#if 0` or unmet `#ifdef` blocks, injecting phantom dependencies and `false` context.
   - **Macro Expansion**: Regex completely fails on computed inclusions like #include MACRO_NAME.
   - **Path Resolution**: Custom regex/path logic lacks knowledge of the compiler search flags (`-I`, `-isystem`) from `compile_commands.json`. Whereas Clang’s internal `HeaderSearch` engine perfectly applies these flags and resolves the exact absolute paths.
   - **Comments and Formatting**: Regex breaks easily on multi-line comments `/* ... */` or line continuations `\` within the directive.
+ Tapping into Clang's built-in Preprocessor delegates all of this to the compiler, completely solving these issues out-of-the-box.

**[Recommendations]**
+ instead of regexs, use Clang's functionality, based on `PPCallbacks` and `InclusionDirective`.
+ It might need additional pre-pass.

ranapratap55:
> I still believe that experimental features should stay experimental for at least one release. Meanwhile, it might be thoroughly tested. Next step: inform end users in the documentation that, with the next release, this particular feature will be enabled by default, and for fallback functionality, here are the options to set explicitly.

Agreed, both `--local-headers` and `--local-headers-recursive` are preserved as explicit opt-in flags in latest revision. 
For the graduation plan,

1. **This release**: Feature ships as opt-in behind `--local-headers/--local-headers-recursive` and document that it will become default. Test it thoroughly.
2. **Next release**: Enable by default with `--skip-local-headers` as fallback for opt-out

ranapratap55:
> **[IMPORTANT]**
> 
> After reviewing this regex-based solution for header tree calculation, I decided it is not a good solution. Using regex in hipify-perl is a forced decision, but **using a regex-based approach with a C++ compiler under the hood can't be treated as a good solution.**
> 
> **[Reasons]**
> 
> * Parsing C/C++ includes with regular expressions is fundamentally brittle and serves as a temporary workaround.
> * Regex has major blind spots that cannot be solved without rewriting a full C++ preprocessor:
>   
>   * **Conditional Compilation**: Regex blindly captures `#include` directives inside `#if 0` or unmet `#ifdef` blocks, injecting phantom dependencies and `false` context.
>   * **Macro Expansion**: Regex completely fails on computed inclusions like #include MACRO_NAME.
>   * **Path Resolution**: Custom regex/path logic lacks knowledge of the compiler search flags (`-I`, `-isystem`) from `compile_commands.json`. Whereas Clang’s internal `HeaderSearch` engine perfectly applies these flags and resolves the exact absolute paths.
>   * **Comments and Formatting**: Regex breaks easily on multi-line comments `/* ... */` or line continuations `\` within the directive.
> * Tapping into Clang's built-in Preprocessor delegates all of this to the compiler, completely solving these issues out-of-the-box.
> 
> **[Recommendations]**
> 
> * instead of regexs, use Clang's functionality, based on `PPCallbacks` and `InclusionDirective`.
> * It might need additional pre-pass.

Thanks for this feedback. I've removed the entire regex-based include scanning system and replaced with Clang's `PPCallbacks::InclusionDirective` in the latest revision.

The new PPCallbacks approach uses `IncludeCollectorCallbacks` (a PPCallbacks subclass), `IncludeCollectorAction` (a PreprocessorFrontendAction), and `collectIncludeTree()` as an entry point

We are directly addressing below concerns:
- Conditional compilation
- Macro expansion
- Path resolution
- Comments/formatting

src/ArgParse.cpp
```diff
@@ -226,46 +226,46 @@ cl::opt<bool> HipDnnSupport("hipdnn",
...
-  std::string(DocRoc.ArgStr),
-  std::string(Experimental.ArgStr),
-  std::string(Versions.ArgStr),
-  std::string(NoUndocumented.ArgStr),
-  std::string(NoWarningsUndocumented.ArgStr),
-  std::string(HipifyAMAP.ArgStr),
-  std::string(HipDnnSupport.ArgStr),
-  std::string(OptLocalHeaders.ArgStr),
-  std::string(OptLocalHeadersRecursive.ArgStr),
+const std::vector<std::string> hipifyOptions{
```
emankov: Please isolate formatting in a separate PR (if formatting is needed).

src/LocalHeader.cpp
```diff
@@ -1,167 +1,291 @@
...
+#if LLVM_VERSION_MAJOR < 19
+                          const clang::Module *SuggestedModule
+#else
+                          const clang::Module *SuggestedModule,
+                          bool ModuleImported
+#endif
+#if LLVM_VERSION_MAJOR > 6
+                          ,
+                          clang::SrcMgr::CharacteristicKind FileType
+#endif
```
emankov, line 89: Might it be simplified?
Did you test it against different LLVM versions?

src/LocalHeader.cpp
```diff
@@ -1,167 +1,291 @@
...
+  explicit IncludeCollectorAction(std::vector<IncludeEntry> &entries)
+      : Entries(entries) {}
 
-  SmallString<256> realBuf;
-  std::error_code ec = llvm::sys::fs::real_path(in, realBuf);
-  if (!ec) return true;
+  void ExecuteAction() override {
+    clang::CompilerInstance &CI = getCompilerInstance();
+    clang::Preprocessor &PP = CI.getPreprocessor();
+    PP.addPPCallbacks(std::make_unique<IncludeCollectorCallbacks>(
```
emankov: For the LLVM backward compatibility, calling `std::make_unique` requires the `<memory>` header.

src/LocalHeader.cpp
```diff
@@ -1,167 +1,291 @@
...
+        for (size_t i = 0; i < Args.size(); ++i) {
+          if (Args[i] == "-include" && i + 1 < Args.size() &&
+              Args[i + 1] == "cuda_runtime.h") {
+            ++i;
+            continue;
+          }
+          filtered.push_back(Args[i]);
+        }
+        return filtered;
+      });
```
emankov: Without `cuda_runtime.h`, Clang lacks base CUDA macros like `__CUDACC__` and `__device__`, causing it to ignore user code hidden behind `#ifdef` blocks.

tests/unit_tests/headers/local_headers/cyclic_header.cu
```diff
@@ -1,4 +1,5 @@
-// RUN: %run_test hipify "%s" "%t" %hipify_args --local-headers-recursive %clang_args
+// RUN: %run_test hipify "%s" "%t" %hipify_args --local-headers-recursive
+// %clang_args
```
emankov: `%clang_args` are an inalienable part of the `RUN:` lit command.

tests/unit_tests/headers/local_headers/injection_recursive.cu
```diff
@@ -0,0 +1,18 @@
+// RUN: %run_test hipify "%s" "%t" %hipify_args --local-headers-recursive
+// %clang_args
```
emankov: `%clang_args` are an inalienable part of the `RUN:` lit command.

tests/unit_tests/headers/local_headers/recursive-header.cu
```diff
@@ -1,4 +1,5 @@
-// RUN: %run_test hipify "%s" "%t" %hipify_args --local-headers-recursive %clang_args
+// RUN: %run_test hipify "%s" "%t" %hipify_args --local-headers-recursive
+// %clang_args
```
emankov: `%clang_args` are an inalienable part of the `RUN:` lit command.

src/LocalHeader.h
```diff
@@ -1,28 +1,70 @@
...
+#include "clang/Tooling/Refactoring.h"
 #include <string>
 #include <vector>
-#include "clang/Tooling/CommonOptionsParser.h"
 
 namespace ct = clang::tooling;
 
+extern bool appendArgumentsAdjusters(ct::RefactoringTool &Tool,
+                                     const std::string &sSourceAbsPath,
+                                     const char *hipify_exe);
```
emankov: Wrong place to `extern`. Actually, `extern` is not needed in `LocalHeader.h`; it is also misleading.
Instead, better to add a forward function declaration in `LocalHeader.cpp`.

tests/unit_tests/headers/local_headers/transitive_system_include.cu
```diff
@@ -0,0 +1,10 @@
+// RUN: %run_test hipify "%s" "%t" %hipify_args --local-headers %clang_args
+
+// CHECK: #include <hip/hip_runtime.h>
+// CHECK-NOT: #include <cuda_runtime.h>
+// CHECK: #include "transitive_parent.h"
+#include <cuda_runtime.h>
+#include <algorithm>
+#include "transitive_parent.h"
```
emankov, line 8: 1. The transitive test does not exercise depth 2, in fact. The option `--local-headers-recursive` should be used instead of `--local-headers`.
2. Add the check for `cudaMalloc` in the `transitive_child.h`.

I anticipate this particular test will fail. In that case, the whole algorithm should be revised and fixed (next iteration).

emankov (changes requested):

src/ArgParse.cpp
ranapratap55: Reverted, No formatting changes in this file.

src/LocalHeader.cpp
ranapratap55: Tested by building HIPIFY against LLVM 14, 15, 16, 18 and all compile without any errors. This covers (`FileEntry*, Optional<FileEntryRef>, OptionalFileEntryRef, and with/without ModuleImported`).

src/LocalHeader.cpp
ranapratap55: Added `<memory>`.

src/LocalHeader.cpp
ranapratap55: `cuda_runtime.h` is filtered from the results via the `isAngled + isWrittenInMainFile` checks.

tests/unit_tests/headers/local_headers/cyclic_header.cu
ranapratap55: Fixed.

tests/unit_tests/headers/local_headers/injection_recursive.cu
ranapratap55: Fixed.

tests/unit_tests/headers/local_headers/recursive-header.cu
ranapratap55: Fixed.

src/LocalHeader.h
ranapratap55: Removed the `extern` and added a local forward declaration of `appendArgumentsAdjusters` in `LocalHeader.cpp` instead.

tests/unit_tests/headers/local_headers/transitive_system_include.cu
ranapratap55: I've switched to `--local-headers-recursive` and added the `cudaMalloc -> hipMalloc` check in `transitive_child.h`. The test now exercises depth-2, discovery finds `transitive_parent.h`, then recursed into `transitive_child.h`. The only fix needed was making `transitive_child.h` self-contained (`#include <algorithm>` for `std::sort`), reason being a recursively-hipified header is compiled standalone.

src/LocalHeader.cpp
ranapratap55: Replaced, no more preceding-include injection ordering.

src/LocalHeader.cpp
```diff
@@ -1,167 +1,280 @@
...
-    if (!readFile(hdr, original)) {
-      errs() << sHipify << sError << "Cannot read header: " << hdr << "\n";
-      continue;
-    }
+    ++current;
 
     std::string hipOut = hdr + ".hip";
     bool ok = hipifySingleSource(hdr, hipOut, compDB, OptionsParserPtr,
-                                  hipify_exe, mainSourceAbsPath, false);
+                                 hipify_exe, mainSourceAbsPath, false);
```
emankov: It looks like a regression. The context vector is never passed to `hipifySingleSource`, and the local headers are being hipified in a vacuum, which should lead to a crash. 

Additionally, the current tests yield `false positives` as they are self-contained. A negative test could help here to prove the injection is actually working.

src/LocalHeader.cpp
```diff
@@ -1,167 +1,280 @@
...
+  // If srcPath has no entry in the compilation database, fall back to a
+  // FixedCompilationDatabase rooted at the mainContextPath's directory so that
+  // the tool doesn't skip the file.
+  std::vector<ct::CompileCommand> cmds = baseDB.getCompileCommands(srcPath);
+  std::unique_ptr<ct::FixedCompilationDatabase> fallbackDB;
+  if (cmds.empty()) {
+    std::string dir = sys::path::parent_path(mainContextPath).str();
+    fallbackDB = std::make_unique<ct::FixedCompilationDatabase>(
+        dir, std::vector<std::string>());
+  }
```
emankov: Passing an empty vector to `FixedCompilationDatabase` wipes out all project compiler flags and breaks header resolution for any real-world project with nested include directories.

src/LocalHeader.cpp
```diff
@@ -1,167 +1,280 @@
...
+                          const clang::Module *SuggestedModule,
+                          bool ModuleImported
+#endif
+#if LLVM_VERSION_MAJOR > 6
+                          ,
+                          clang::SrcMgr::CharacteristicKind FileType
+#endif
+                          ) override {
+    if (!SM.isWrittenInMainFile(hash_loc))
+      return;
```
emankov: Using `!SM.isWrittenInMainFile(hash_loc)` inside the preprocessor callback forcibly prevents `Clang` from natively traversing nested includes and forces the tool to spawn a brand-new compiler instance for every child header, which destroys macro inheritance. The possible solution here is to remove the check.

src/LocalHeader.cpp
```diff
@@ -1,167 +1,280 @@
...
-  SmallString<256> norm = in;
-  llvm::sys::path::remove_dots(norm, true);
-  return llvm::sys::fs::exists(norm);
-}
+    PP.EnterMainSourceFile();
+    clang::Token Tok;
+    do {
+      PP.Lex(Tok);
+    } while (Tok.isNot(clang::tok::eof));
+  }
```
emankov: Overriding `ExecuteAction()` to manually call `EnterMainSourceFile()` and `Lex` bypasses standard Clang initialization, such as `IgnorePragmas()`. If a header contains unknown pragmas, your pre-pass might crash.

src/LocalHeader.cpp
```diff
@@ -1,167 +1,280 @@
...
+#if LLVM_VERSION_MAJOR < 19
+                          const clang::Module *SuggestedModule
+#else
+                          const clang::Module *SuggestedModule,
+                          bool ModuleImported
+#endif
+#if LLVM_VERSION_MAJOR > 6
+                          ,
+                          clang::SrcMgr::CharacteristicKind FileType
+#endif
```
emankov: Several parameters in `InclusionDirective` signature are named but never actually used in the function body.

emankov (changes requested):

src/LocalHeader.cpp
ranapratap55: Fixed in latest code.

src/LocalHeader.cpp
ranapratap55: Thanks for this. I've removed the check now and `includerPath`, `isFromMainFile`, `isSystem` are modified as data, so one preprocessor run gives the whole tree.  The `recursive/non-recursive` split is a single filter. Removal alone wasn't ideal though, since CUDA's headers use quoted includes. So paired with `SM.isInSystemHeader(HashLoc)` and an empty includer path check.

src/LocalHeader.cpp
ranapratap55: Fixed this. `IncludeCollectorAction` derives from `PreprocessOnlyAction` and registers the callbacks in `BeginSourceFileAction()`, so the lexing loop and `IgnorePragmas()` come from the framework. The old loop was the same code without `IgnorePragmas()`, so unknown pragmas warned. Collection now propagates `Tool.run()` failure, a `-Werror` build would abort on something like `#pragma unroll`. Known pragmas such as` #pragma once` are untouched.

src/LocalHeader.cpp
ranapratap55: Removed the branch entirely and it is unreachable now. Without `-p` we hold a `FixedCompilationDatabase`, which synthesizes a command for any path. With `-p`, `loadFromDirectory` wraps the JSON database in `inferMissingCompileCommands`, which borrows the command of the closest matching file, swaps in the filename and drops output args. `hipifySingleSource` already relies on this, passing a `/tmp/*.hip` path that's in no database.

src/LocalHeader.cpp
ranapratap55: `buildIncludeContext()` now derives it from the entries of that same preprocessor run. The ancestor chain's directives that precede the header, oldest first. Ancestors themselves are excluded, since reincluding a parent pulls the header back through its own guard and empties the copy being hipified. The injection also moved from `BEGIN` to a single `END` insertion, so it follows `-include cuda_runtime.h`. 

Added a negative test for non self-contained headers. `nsc_context.h` declares `NscVec` and is self-contained,`nsc_dependent.h` uses `NscVec` and `cudaMemset` while including nothing. `non_self_contained.cu` includes them in that order under `--local-headers`. Because `-include cuda_runtime.h` is always injected, `cudaError_t` and `cudaMemset` resolve on their own. So `NscVec` is the only symbol that can come from nowhere but the injected context. With an empty vector the header fails with `unknown type name NscVec`, hipify returns non-zero and the test fails. `nsc_dependent.h` sits in `config.excludes` because it cannot carry a `RUN`.

src/LocalHeader.cpp
```diff
@@ -1,167 +1,283 @@
...
-  static const std::regex LocalIncludeRe(
-      R"(^\s*#\s*include\s*\"([^\"\n]+)\"\s*(?://.*)?$)", std::regex::ECMAScript);
+    IncludeEntry Entry;
+    Entry.fileName = FileName.str();
+    Entry.isAngled = IsAngled;
+    Entry.includerPath = std::move(IncluderPath);
+    Entry.isFromMainFile = IncluderID == SM.getMainFileID();
+#if LLVM_VERSION_MAJOR > 6
+    Entry.isSystem = FileType != clang::SrcMgr::C_User;
+#endif
```
emankov, line 103:
```cpp
#else
    Entry.isSystem = IsAngled;
```

src/LocalHeader.cpp
```diff
@@ -1,167 +1,283 @@
...
-bool hipifyLocalHeaders(const std::string &mainSourceAbsPath,
-                             const ct::CompilationDatabase *compDB,
-                             ct::CommonOptionsParser *OptionsParserPtr,
-                             const char *hipify_exe,
-                             bool recursive) {
+} // namespace
+
+bool appendArgumentsAdjusters(ct::RefactoringTool &Tool,
+                              const std::string &sSourceAbsPath,
+                              const char *hipify_exe);
```
emankov, line 207: Better to move this declaration into `LocalHeader.h` with `extern`.

src/LocalHeader.cpp
```diff
@@ -1,167 +1,283 @@
...
+
+bool collectIncludeTree(const std::string &srcPath,
+                        const ct::CompilationDatabase *compDB,
+                        ct::CommonOptionsParser *OptionsParserPtr,
+                        const char *hipify_exe,
+                        std::vector<IncludeEntry> &outEntries) {
+  outEntries.clear();
+
+  ct::RefactoringTool Tool(
+      compDB ? *compDB : OptionsParserPtr->getCompilations(), {srcPath});
```
emankov, line 217: We need to revise this explicit `JSON` compilation database detection here and elsewhere, since the `getCompilations()` database performs it implicitly.

**[Explanation]**
**Situation 1**: The workspace has a valid compile_commands.json. When a user runs HIPIFY in a project where the build system (like CMake) has generated a compile_commands.json file, the compDB variable becomes valid (not null). The code initializes the tool using a ternary operator: `compDB ? *compDB : OptionsParserPtr->getCompilations()`. Because of this operator, the tool strictly prioritizes the raw database (*compDB) and completely bypasses `OptionsParserPtr->getCompilations()`.

**Situation 2**: The tool attempts to process a local header file (`.h`, or `.hpp`, or `cuh`). By their very nature, compilation databases (compile_commands.json) only record compile commands for translation units (like `.cu`, `.cpp`, or `.c` files). Header files are never recorded in this database because they are not compiled independently.

**The Mechanism of Failure**:

1. HIPIFY discovers a local header and attempts to hipify it.
2. The tool queries the raw *compDB database, asking for the specific compile flags for that .h file.
3. Because the header file is not listed in the database, the raw database returns nothing.
4. Normally, the smart wrapper (CommonOptionsParser) would act as a fallback safety net: it automatically overlays global command-line arguments (like manually passed -I flags) to generate a compile command on the fly for missing files.
5. However, because the code forces the use of the raw *compDB, it bypasses this built-in safety net.
6. Consequently, `ClangTool::run()` finds no command, outputs `"Skipping... Compile command not found"`, and silently aborts.

emankov (changes requested):
Thank you. I do not see any other issues

src/LocalHeader.cpp
ranapratap55: Updated, thanks.

src/LocalHeader.cpp
ranapratap55: Updated.

src/LocalHeader.cpp
ranapratap55: I debugged more into this, and I do not think failure mode is reachable. `compDB` is not a raw database. `main.cpp:393` calls `CompilationDatabase::loadFromDirectory` then reaches `JSONCompilationDatabasePlugin::loadFromDirectory` and returns `inferTargetAndDriverMode(inferMissingCompileCommands((...)))`. The comment at line 159 says it also `infers compile commands for files not present in the database`. So the header inference you're describing is already inside `*compDB`.

`getCompilations()` is that same object plus `ArgumentsAdjustingCompilations`, which rewrites commands the inner database already returned and so cannot synthesize one for a missing file. Our `/unit_tests/compilation_database/` tests show it is working. They run with `-p=` and hipify a `/tmp/*.hip` path that is in no JSON. That is the interpolating wrapper answering through `*compDB`.

You are right that the ternary redundant. The `compDB` branch drops `--extra-arg`, and the JSON is parsed twice per run. So I'd rather fix it as a separate NFC PR, with a `-p` plus `--local-headers` test as the first commit, since no combination covers it as of today.

kzhuravl (approved):

PR #2406, opened 2026-02-23
[HIPIFY][fix] `Stack Overflow` possible Issues - Final clean up and formatting

emankov:
+ Removed redundant `()` from the lambda initializations
+ [Reason]
  - Omitting the redundant parentheses reduces visual noise and makes it much clearer that this is an Immediately Invoked Function Expression (`IIFE`), allowing the trailing `()` execution call at the end of the block to stand out
+ Formatting


ranapratap55 (approved):
LGTM.

Issue #2062, opened 2025-08-14
[HIPIFY][feature] Partial hipification support for a particular CUDA API


emankov:
**[Synopsis]**
+ We already have situations with changed API/ABI from both `CUDA` and `HIP`.
+ For example, the `cuCtxCreate` API has changed in `CUDA 13.0.0`, and it is currently unsupported by `HIP`; but it is supported for the previous CUDA versions.
+ For those supported CUDA versions, such an API should have an option for attributing the supported CUDA versions.
+ Hipification of such a partially supported API should go with or without warning.
+ Currently, if a particular CUDA API is marked as `HIP_UNSUPPORTED`, hipification doesn't occur despite the `CUDA_VERSION`.

**[Tasks]**
+ Additional API attributing for specifying the supported `CUDA` versions
   - [Short Term] Attribute for the latest supported `CUDA` version
   - [Long Term] Ability to specify version lists and ranges
+ Hipify partially supported API for the supported `CUDA` version(s)
+ Do not hipify partially supported API for the unsupported `CUDA` version(s)
+ [doc] Additional column for indicating such partially supported APIs
+ [hipify-perl] Update correspondingly

Issue #837, opened 2023-04-13
[HIPIFY][feature] Add a new function call transformation type "additional non-const arg"


emankov:
The additional arg should be declared before the function call with the specified assigned value or created in case of pointer type.

**[Example]**
cudnnGetRNNDescriptor_v6 -> miopenGetRNNDescriptor_V2,
where such an additional arg is `miopenRNNBiasMode_t *biasMode`

emankov:
Implemented by #1001

PR #2415, opened 2026-03-02
[HIPIFY][#837][#2062][RFC][feature] Implement `e_insert_new_argument` for APIs need variable declarations - Part 1

ranapratap55:
## Motivation

Issue #837, additional arguments must be declared before the function call. This is necessary because several cuDNN to MIOpen mappings require parameters such as `use_mask, state_evo, and rng_mode` for `miopenSetDropoutDescriptor` which lack CUDA equivalents and cannot be inferred from the current context. 

## Problem Statement

When hipify-clang encounters a CUDA API whose HIP/MIOpen equivalent has additional arguments that are not present in the original call, it currently has no mechanism to
1. Declare new variables before the function call
2. Append new variables as arguments after hipified call
3. Warn user that new variables need initialization before usage

## Solution

Introduces `e_insert_new_argument`:
1. Declares a new variable before the function call 
2. Appends the variable as an argument
3. Emits a warning to user for each inserted variable about intialization

## Summary of changes in this PR:

- `e_insert_new_argument` type for APIs where MIOpen requires additional arguments not present in CUDA #837
- Variable declarations generated with value-initialization (`= {}`) and `hipify_` prefix
- Middle insertions at correct positions, end-appends batched to work for same-offset constraint
- Compiler warning emitted for each generated variable instructing user to initialize appropriately

**ToDo work (follow-up PRs):**

- [ ] **Braceless control flow scope**: When API call is inside a single-statement `if/for/while` without braces, the inserted declarations break control flow. **Fix**: query AST parents to detect `CompoundStmt` and wrap with `{ }`.
- [ ] **Multiple-declaration guard**: If the same API is called twice in same scope, the `hipify_` variables would be declared twice. **Fix**: append a number (e.g., `hipify_use_mask_1`).

### DIFF
--- a/src/CUDA2HIP_Scripting.h
+++ b/src/CUDA2HIP_Scripting.h
@@ -41,6 +41,7 @@ namespace hipify {
     e_add_var_argument,
     e_move_argument,
     e_replace_argument_with_const,
+    e_insert_new_argument,
   };
 
   enum OverloadTypes {
@@ -50,6 +51,7 @@ namespace hipify {
   enum CastWarning {
     cw_None,
     cw_DataLoss,
+    cw_NeedsNewArgDecl,
   };
 
   enum OverloadWarning {
@@ -62,6 +64,8 @@ namespace hipify {
     std::string constValToAddOrReplace = "";
     unsigned moveOrCopyTo = 0;
     unsigned numberToMoveOrCopy = 1;
+    std::string newArgTypeName = "";
+    bool isPointerArg = false;
   };
 
   typedef std::map<unsigned, CastInfo> ArgCastMap;
--- a/src/CUDA2HIP_Scripting.h
+++ b/src/CUDA2HIP_Scripting.h
@@ -61,11 +61,14 @@ namespace hipify {
   struct CastInfo {
     CastTypes castType = e_HIP_SYMBOL;
     CastWarning castWarn = cw_None;
+    // For e_add_const_argument / e_replace_argument_with_const: the literal value to insert or replace.
+    // For e_insert_new_argument: the variable name to declare and pass as an argument.
     std::string constValToAddOrReplace = "";
     unsigned moveOrCopyTo = 0;
     unsigned numberToMoveOrCopy = 1;
     std::string newArgTypeName = "";
     bool isPointerArg = false;
+    std::string defaultInitValue = "";
   };
 
   typedef std::map<unsigned, CastInfo> ArgCastMap;
--- a/src/HipifyAction.cpp
+++ b/src/HipifyAction.cpp
@@ -599,7 +599,7 @@ std::map<std::string, std::vector<ArgCastStruct>> FuncArgCasts {
       {
         {
           {0, {e_remove_argument, cw_None}},
-          {7, {e_insert_new_argument, cw_NeedsNewArgDecl, "biasMode", 7, 1, "miopenRNNBiasMode_t", true}}
+          {8, {e_insert_new_argument, cw_NeedsNewArgDecl, "biasMode", 0, 1, "miopenRNNBiasMode_t", true}}
         },
         true,
         true
@@ -611,7 +611,7 @@ std::map<std::string, std::vector<ArgCastStruct>> FuncArgCasts {
       {
         {
           {0, {e_remove_argument, cw_None}},
-          {7, {e_insert_new_argument, cw_NeedsNewArgDecl, "biasMode", 7, 1, "miopenRNNBiasMode_t", false}}
+          {8, {e_insert_new_argument, cw_NeedsNewArgDecl, "biasMode", 0, 1, "miopenRNNBiasMode_t", false}}
         },
         true,
         true
@@ -2733,43 +2733,47 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
       std::string combinedArgText;
       bool hasInsertions = false;
       std::vector<hipify::CastInfo> insertionWarnings;
+      clang::SourceLocation callBeginLoc = call->getBeginLoc();
+      unsigned callCol = SM.getSpellingColumnNumber(callBeginLoc);
+      clang::SourceLocation stmtInsertLoc = callBeginLoc.getLocWithOffset(-static_cast<int>(callCol - 1));
       std::string indent;
-      {
-        clang::SourceLocation loc = call->getBeginLoc();
-        unsigned col = SM.getSpellingColumnNumber(loc);
-        clang::SourceLocation lineStart = loc.getLocWithOffset(-static_cast<int>(col - 1));
-        const char *p = SM.getCharacterData(lineStart);
-        while (*p == ' ' || *p == '\t')
-          indent += *p++;
-      }
+      const char *indentPtr = SM.getCharacterData(stmtInsertLoc);
+      while (*indentPtr == ' ' || *indentPtr == '\t')
+        indent += *indentPtr++;
       for (auto c : cc.castMap) {
         if (c.second.castType == e_insert_new_argument) {
           hasInsertions = true;
-          combinedDeclText += indent + c.second.newArgTypeName + " " + c.second.constValToAddOrReplace + ";\n";
+          std::string initExpr = c.second.defaultInitValue.empty() ? "{}" : c.second.defaultInitValue;
+          combinedDeclText += indent + c.second.newArgTypeName + " " + c.second.constValToAddOrReplace + " = " + initExpr + ";\n";
           std::string argText;
           if (c.second.isPointerArg)
             argText = "&" + c.second.constValToAddOrReplace;
           else
             argText = c.second.constValToAddOrReplace;
-          if (!combinedArgText.empty())
-            combinedArgText += ", ";
-          combinedArgText += argText;
+          if (c.first < call->getNumArgs()) {
+            clang::SourceLocation argLoc = call->getArg(c.first)->getBeginLoc();
+            ct::Replacement middleRep(SM, argLoc, 0, argText + ", ");
+            clang::FullSourceLoc middleFullSL(argLoc, SM);
+            insertReplacement(middleRep, middleFullSL);
+          } else {
+            if (!combinedArgText.empty())
+              combinedArgText += ", ";
+            combinedArgText += argText;
+          }
           insertionWarnings.push_back(c.second);
         }
       }
       if (hasInsertions) {
-        clang::SourceLocation callBeginLoc = call->getBeginLoc();
-        unsigned col = SM.getSpellingColumnNumber(callBeginLoc);
-        clang::SourceLocation stmtInsertLoc = callBeginLoc.getLocWithOffset(-static_cast<int>(col - 1));
         ct::Replacement declRep(SM, stmtInsertLoc, 0, combinedDeclText);
         clang::FullSourceLoc declFullSL(stmtInsertLoc, SM);
         insertReplacement(declRep, declFullSL);
-
-        clang::SourceLocation insertLoc = call->getEndLoc();
-        std::string fullArgText = ", " + combinedArgText;
-        ct::Replacement argRep(SM, insertLoc, 0, fullArgText);
-        clang::FullSourceLoc argFullSL(insertLoc, SM);
-        insertReplacement(argRep, argFullSL);
+        if (!combinedArgText.empty()) {
+          clang::SourceLocation insertLoc = call->getEndLoc();
+          std::string fullArgText = ", " + combinedArgText;
+          ct::Replacement argRep(SM, insertLoc, 0, fullArgText);
+          clang::FullSourceLoc argFullSL(insertLoc, SM);
+          insertReplacement(argRep, argFullSL);
+        }
 
         for (auto &info : insertionWarnings) {
           clang::DiagnosticsEngine &DE = getCompilerInstance().getDiagnostics();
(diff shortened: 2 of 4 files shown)


src/HipifyAction.cpp
```diff
@@ -2712,7 +2729,61 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
...
+        clang::SourceLocation callBeginLoc = call->getBeginLoc();
+        unsigned col = SM.getSpellingColumnNumber(callBeginLoc);
+        clang::SourceLocation stmtInsertLoc = callBeginLoc.getLocWithOffset(-static_cast<int>(col - 1));
+        ct::Replacement declRep(SM, stmtInsertLoc, 0, combinedDeclText);
+        clang::FullSourceLoc declFullSL(stmtInsertLoc, SM);
+        insertReplacement(declRep, declFullSL);
+
+        clang::SourceLocation insertLoc = call->getEndLoc();
+        std::string fullArgText = ", " + combinedArgText;
+        ct::Replacement argRep(SM, insertLoc, 0, fullArgText);
```
emankov: The current implementation assumes that all newly inserted arguments should be appended to the end of the function call.

This breaks API signatures in which the new argument should be placed in the middle of the parameter list. For example, in the patch's tests for `cudnnSetRNNDescriptor_v6` to `miopenSetRNNDescriptor_V2`.

src/HipifyAction.cpp
```diff
@@ -2712,7 +2729,61 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
...
+      std::vector<hipify::CastInfo> insertionWarnings;
+      std::string indent;
+      {
+        clang::SourceLocation loc = call->getBeginLoc();
+        unsigned col = SM.getSpellingColumnNumber(loc);
+        clang::SourceLocation lineStart = loc.getLocWithOffset(-static_cast<int>(col - 1));
+        const char *p = SM.getCharacterData(lineStart);
+        while (*p == ' ' || *p == '\t')
+          indent += *p++;
+      }
```
emankov: Not quite understand such a scoping.

src/HipifyAction.cpp
```diff
@@ -607,7 +610,21 @@ std::map<std::string, std::vector<ArgCastStruct>> FuncArgCasts {
...
+    }
+  },
+  {sCudnnSetDropoutDescriptor,
+    {
+      {
+        {
+          {6, {e_insert_new_argument, cw_NeedsNewArgDecl, "use_mask", 6, 1, "bool", false}},
+          {7, {e_insert_new_argument, cw_NeedsNewArgDecl, "state_evo", 7, 1, "bool", false}},
+          {8, {e_insert_new_argument, cw_NeedsNewArgDecl, "rng_mode", 8, 1, "miopenRNGType_t", false}}
```
emankov: I think the declarations should be generated as initialized variables. Is it so in fact?

src/HipifyAction.cpp
```diff
@@ -2712,7 +2729,61 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
...
+            combinedArgText += ", ";
+          combinedArgText += argText;
+          insertionWarnings.push_back(c.second);
+        }
+      }
+      if (hasInsertions) {
+        clang::SourceLocation callBeginLoc = call->getBeginLoc();
+        unsigned col = SM.getSpellingColumnNumber(callBeginLoc);
+        clang::SourceLocation stmtInsertLoc = callBeginLoc.getLocWithOffset(-static_cast<int>(col - 1));
+        ct::Replacement declRep(SM, stmtInsertLoc, 0, combinedDeclText);
```
emankov: The variable declaration logic determines the line start and inserts the new variables directly above the function call.

If the original function call is inside a single-statement control flow block without braces, this textual insertion will break the AST logic.

For instance:
```cpp
if (status == CUDNN_STATUS_SUCCESS)
    cudnnSetDropoutDescriptor(...);
```
will be hipified to smth. like the following:

```cpp
if (status == CUDNN_STATUS_SUCCESS)
bool use_mask;
bool state_evo;
miopenRNGType_t rng_mode;
    miopenSetDropoutDescriptor(...); // This is now outside the if-statement
```

**[Recommendation]**

To make this robust, check if the parent AST node is a `CompoundStmt`. If it's not (e.g., it's an `IfStmt` or `ForStmt`), you should generate `{ and }` around the inserted declarations and the function call to preserve the original control flow scope.

src/HipifyAction.cpp
emankov: The current implementation assumes that all newly inserted arguments should be appended to the end of the function call.

This breaks API signatures that require the new argument to be placed in the middle of the parameter list. For example, in the patch's tests for `cudnnSetRNNDescriptor_v6` to `miopenSetRNNDescriptor_V2`.

**[Recommendation]**

Instead of handling e_insert_new_argument in an isolated pre-pass and skipping it in the main argument iteration loop `(if (c.second.castType == e_insert_new_argument) continue;)`, you should integrate the insertion directly into the existing argument parsing loop. This allows you to use `c.first` (the argument index) to correctly interleave the new arguments with the existing `Expr*` arguments.

src/HipifyAction.cpp
```diff
@@ -596,7 +598,8 @@ std::map<std::string, std::vector<ArgCastStruct>> FuncArgCasts {
     {
       {
         {
-          {0, {e_remove_argument, cw_None}}
+          {0, {e_remove_argument, cw_None}},
+          {7, {e_insert_new_argument, cw_NeedsNewArgDecl, "biasMode", 7, 1, "miopenRNNBiasMode_t", true}}
```
emankov: By using hardcoded names like `biasMode` and `use_mask`, the tool risks causing variable redefinition errors if the user's codebase already uses those variable names in the local scope.

**[Recommendation]**

Consider prefixing the inserted variables to guarantee uniqueness (e.g., `hipify_use_mask`, `hipify_biasMode`, or with `hip_` prefix) or appending a counter.

src/CUDA2HIP_Scripting.h
```diff
@@ -62,6 +64,8 @@ namespace hipify {
     std::string constValToAddOrReplace = "";
```
emankov, line 66: [Minor]

I'd rename `constValToAddOrReplace` to smth. like `newArgName` and place it before `std::string newArgTypeName = "";`.

emankov (changes requested):
Good try on an important feature from our ToDo list, but see my comments, please.

src/HipifyAction.cpp
ranapratap55: yes, scoping is not required here. 
Refactored the code, which eliminates both scoping block and duplicate computation for indentation and declaration.

src/HipifyAction.cpp
ranapratap55: yes, #837 says "declared before the function call with the specified assigned value."

Added a `defaultInitValue` field to `CastInfo` which initializes to specified value or zero initialization ( `= {}` ).

src/HipifyAction.cpp
ranapratap55: Correct, this is a real issue here. 

I can implement the AST check in this current PR. It is not only to `e_insert_new_argument` but it applies to future feature that inserts statements above a call.

it can be designed as a reusable utility with multiple control flow variants like (`if, else, for, while, do-while`).

Would you prefer I handle it here, or as a dedicated follow-up PR?

src/HipifyAction.cpp
ranapratap55: name collisions are possible. I intentionally chose `biasMode`, `use_mask` to match the `MIOpen` API parameter names. We are already warning users to review and initialize these variables. 

If you feel strongly about `hip_`, I can add this change happily.

src/CUDA2HIP_Scripting.h
ranapratap55: `constValToAddOrReplace` is an existing field shared by multiple cast types.

- For `e_add_const_argument` - a literal constant value (e.g., "miopenLRNCrossChannel", "true", "hipHostMallocDefault")
- for `e_replace_argument_with_const` - a replacement value (e.g., "rocsparse_solve_policy_auto")
- for `e_insert_new_argument` - a variable name.

Renaming to `newArgName` would be misleading for the first two cases. 

I've added a clarifying comment to the field.

src/HipifyAction.cpp
ranapratap55: Thanks, the old logic always appends at the end. I have added the logic for middle insertion in updated code.

The reason I used a separate pass instead of merging into the main argument loop is LLVM constraint of `Replacements::add()` which rejects multiple insertions at the same source offset. For example, `cudnnSetDropoutDescriptor`, which has 6 CUDA args. For MIOpen args, insertion takes place at positions 6, 7, 8 that is beyond the end. In the source text, all three new insertions at the same location `)` at `call->getEndLoc()`. So `end appends` batched into a single replacement, we concatenate them as `, use_mask, state_evo, rng_mode` and then emits one replacement.

For middle insertion, each middle insertion targets a unique source offset, For an API needed multiple middle insertions, each will go to a different offset.

src/HipifyAction.cpp
emankov: I think we need to consider the naming and syntax of the code generated by `hipify-clang`. I do not think strongly about a particular prefix for variables and constants to appear after hipification. But what about `hipify_` for new variables and consts?

@kzhuravl, @searlmc1,
Does it make sense, or is there a more suitable solution? Anyway, we need to point out the generated variables (and functions in the future), so their naming differs from CUDA and, I suppose, HIP naming, and we should check for multiple declarations or reuse.

src/HipifyAction.cpp
```diff
@@ -2712,7 +2729,65 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
       if (TranslateToRoc == true && cc.isToRoc == false) continue;
       if (TranslateToRoc == false && cc.isToRoc == true && TranslateToMIOpen == false) continue;
       clang::LangOptions DefaultLangOptions;
+      std::string combinedDeclText;
+      std::string combinedArgText;
+      bool hasInsertions = false;
+      std::vector<hipify::CastInfo> insertionWarnings;
+      clang::SourceLocation callBeginLoc = call->getBeginLoc();
+      unsigned callCol = SM.getSpellingColumnNumber(callBeginLoc);
+      clang::SourceLocation stmtInsertLoc = callBeginLoc.getLocWithOffset(-static_cast<int>(callCol - 1));
```
emankov, line 2738: **[Control Flow Scoping]**
While the commit message mentions "`Fix scoping`, the code still relies on a raw textual offset to find the beginning of the line. Because this is a purely textual insertion at the indentation level, it will still break single-statement control flow blocks. For example:

```cpp
if (status == CUDNN_STATUS_SUCCESS)
    status = cudnnSetDropoutDescriptor(...);
```

will be hipified to:
```cpp
if (status == CUDNN_STATUS_SUCCESS)
    bool use_mask = {};
    bool state_evo = {};
    miopenRNGType_t rng_mode = {};
    status = miopenSetDropoutDescriptor(...); // Danger: This is no longer guarded by the 'if'!
```

The same goes for tests. They should cover different control flow cases.

**[Posible Solution]**
To query the `Clang AST` to see if the current statement is a direct child of a control flow node (like `IfStmt`, `ForStmt`, `WhileStmt`) without a `CompoundStmt` (braces) in between, or even better, to find a more common, higher-level solution than iterating through control flow node types.

emankov (changes requested):
The middle argument insertion is a very elegant solution!

Using `call->getArg(c.first)->getBeginLoc()` to seamlessly interleave the new arguments into the existing Clang replacement stream is much cleaner than reconstructing the entire argument list. This is the approach we prefer to use when dealing with arguments in Clang. Adding the `{}` default initialization is also a great touch to squash uninitialized variable warnings right out of the gate.

However, we need to address the scoping implementation, as _it didn't fully resolve the control-flow vulnerability_.

src/HipifyAction.cpp
ranapratap55: Agreed. I've updated all generated variable names with `hipify_` prefix. 

If the same API is called twice in the same scope, the `hipify_` variables would be declared twice. I will address this issue by appending a number (e.g., `hipify_use_mask_1`) in a folow-up PR.

src/HipifyAction.cpp
ranapratap55: To clarify the "Fix scoping" in my earlier commit was for
> Not quite understand such a scoping. https://github.com/ROCm/HIPIFY/pull/2415#discussion_r2879210321

the anonymous scoping block review comment, not for the braceless control flow issue.

The control flow fix (AST parent) applies beyond `e_insert_new_argument`. Any future feature that inserts statements above a API call would need the same utility. I'd prefer to handle it in a dedicated follow-up PR.

ranapratap55:
> The middle argument insertion is a very elegant solution!
> 
> Using `call->getArg(c.first)->getBeginLoc()` to seamlessly interleave the new arguments into the existing Clang replacement stream is much cleaner than reconstructing the entire argument list. This is the approach we prefer to use when dealing with arguments in Clang. Adding the `{}` default initialization is also a great touch to squash uninitialized variable warnings right out of the gate.
> 
> However, we need to address the scoping implementation, as _it didn't fully resolve the control-flow vulnerability_.

Marking this PR as partial and will create a ToDo for the below 

1. Braceless control flow (AST parent)
2. Multiple declaration guard

src/HipifyAction.cpp
emankov: As follow-up PRs

emankov (approved):
@ranapratap55,

Thank you for stepping into this new `generative era` of `hipify-clang`.

`That's one small step for an RFC, one giant leap for HIPIFY.` 
Btw, I'm rewording Armstrong's famous phrase without a shade of sarcasm.

Good job, go ahead with this RFC and those to come.

PR #2428, opened 2026-03-16
[HIPIFY][Perl][#1776][fix] Fixed `CONVERTED refs count` calculation under the `-print-stats`

emankov:
**[Bug]**
+ In `CUDA2HIP_Perl.cpp`, the variable `$hasDeviceCode` is assigned before the line-scanning loops for device functions and types

**[Solution]**
+ Move the `$hasDeviceCode` assignment to strictly after the `foreach` scanning loops

**[Side Effect]**
+ Minor `hipify-perl` performance improvement in the amount of statistical error

### DIFF
--- a/bin/hipify-perl
+++ b/bin/hipify-perl
@@ -18126,14 +18126,6 @@ while (@ARGV) {
                                 $warnings++;
                             }
                         }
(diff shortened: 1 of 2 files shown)


ranapratap55 (approved):
LGTM.

PR #2434, opened 2026-03-18
[HIPIFY][Perl][#181][fix] Fix incomplete directory traversal for `-exclude-dirs`

emankov:
**[Synopsis]**
+ Previously, the `-exclude-dirs` mechanism in the generated `hipify-perl` evaluated only the immediate parent directory of a given file.
+ Consequently, the script failed to skip files (e.g., hipified them) buried deep inside excluded folders because it only checked their immediate parent directory.

**[Solution]**
+ Updated the C++ generator to implement full directory tree traversal for the `-exclude-dirs` flag provided.
+ The generated script now iteratively checks all subdirectories, guaranteeing that deeply nested files within an excluded directory's subtree are correctly skipped and not hipified.
+ This PR fixes the feature (#181) implemented by [Balint Joo](https://github.com/bjoo) (NVIDIA Corporation).


ranapratap55 (approved):
LGTM.

PR #2443, opened 2026-03-24
[HIPIFY][#837][#2062][#2415] Add multi-declaration gaurd check for `e_insert_new_argument` - Part 2

ranapratap55:
## Context

Follow-up PR to #2415 for `e_insert_new_argument` cast type that generates variable declarations above a function call when hipifying APIs where MIOpen requires additional arguments not present in CUDA.

One known limitation in #2415 was:

> **Multiple-declaration guard**: If the same API is called twice in the same scope,
> the `hipify_` variables would be declared twice, causing a compile-time redefinition
> error.

With this PR, we are fixing it.

## Problem

Given a CUDA code with the same API called multiple times in the same scope:

```cpp
status = cudnnSetDropoutDescriptor(desc, handle, dropout, states, size, seed);
status = cudnnSetDropoutDescriptor(desc, handle, dropout, states, size, seed);
```
#2415 would generate 
```cpp
bool hipify_use_mask = {};       // works
bool hipify_use_mask = {};       // fails — REDEFINITION ERROR
```
## Solution

Added `InsertedVarCounter` that tracks the count for each base variable name that has been generated.

Also fixed warning emission loop with `insertionVarNames` vector so the compiler warning message correctly reports the unique name `hipify_use_mask_1` instead of the base name `hipify_use_mask`.

With this change code would generate like:
```cpp
bool hipify_use_mask = {};         // works
bool hipify_use_mask_1 = {};       // works - unique name
bool hipify_use_mask_2 = {};       // works - unique name
```

### DIFF
--- a/src/HipifyAction.cpp
+++ b/src/HipifyAction.cpp
@@ -2732,8 +2732,7 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
       std::string combinedDeclText;
       std::string combinedArgText;
       bool hasInsertions = false;
-      std::vector<hipify::CastInfo> insertionWarnings;
-      std::vector<std::string> insertionVarNames;
+      std::vector<std::pair<hipify::CastInfo, std::string>> insertions;
       clang::SourceLocation callBeginLoc = call->getBeginLoc();
       unsigned callCol = SM.getSpellingColumnNumber(callBeginLoc);
       clang::SourceLocation stmtInsertLoc = callBeginLoc.getLocWithOffset(-static_cast<int>(callCol - 1));
@@ -2744,7 +2743,8 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
       for (auto c : cc.castMap) {
         if (c.second.castType == e_insert_new_argument) {
           const std::string &baseName = c.second.constValToAddOrReplace;
-          unsigned counter = InsertedVarCounter[baseName]++;
+          clang::FileID fileID = SM.getFileID(call->getBeginLoc());
+          unsigned counter = InsertedVarCounter[{fileID, baseName}]++;
           std::string uniqueName = (counter == 0) ? baseName : baseName + "_" + std::to_string(counter);
           hasInsertions = true;
           std::string initExpr = c.second.defaultInitValue.empty() ? "{}" : c.second.defaultInitValue;
@@ -2764,8 +2764,7 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
               combinedArgText += ", ";
             combinedArgText += argText;
           }
-          insertionWarnings.push_back(c.second);
-          insertionVarNames.push_back(uniqueName);
+          insertions.push_back({c.second, uniqueName});
         }
       }
       if (hasInsertions) {
@@ -2779,15 +2778,15 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
           clang::FullSourceLoc argFullSL(insertLoc, SM);
           insertReplacement(argRep, argFullSL);
         }
-
-        for (size_t i = 0; i < insertionWarnings.size(); ++i) {
+        for (auto &ins : insertions) {
           clang::DiagnosticsEngine &DE = getCompilerInstance().getDiagnostics();
           const auto ID = DE.getCustomDiagID(clang::DiagnosticsEngine::Warning,
             "HIP API '%0' requires additional argument '%1' of type '%2'. "
             "A variable declaration has been inserted before the call. "
             "Please initialize it appropriately.");
           clang::FullSourceLoc warnFullSL(call->getBeginLoc(), SM);
-          DE.Report(warnFullSL, ID) << sName << insertionVarNames[i] << insertionWarnings[i].newArgTypeName;
+          DE.Report(warnFullSL, ID)
+              << sName << ins.second << ins.first.newArgTypeName;
         }
       }
       for (auto c : cc.castMap) {
--- a/src/HipifyAction.cpp
+++ b/src/HipifyAction.cpp
@@ -2733,6 +2733,7 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
       std::string combinedArgText;
       bool hasInsertions = false;
       std::vector<hipify::CastInfo> insertionWarnings;
+      std::vector<std::string> insertionVarNames;
       clang::SourceLocation callBeginLoc = call->getBeginLoc();
       unsigned callCol = SM.getSpellingColumnNumber(callBeginLoc);
       clang::SourceLocation stmtInsertLoc = callBeginLoc.getLocWithOffset(-static_cast<int>(callCol - 1));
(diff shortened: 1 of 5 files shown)


src/HipifyAction.cpp
```diff
@@ -2775,14 +2780,14 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
           insertReplacement(argRep, argFullSL);
         }
 
-        for (auto &info : insertionWarnings) {
+        for (size_t i = 0; i < insertionWarnings.size(); ++i) {
```
emankov: [Minor] 
We can keep the former "for loop" declaration with `auto`, based even on the `C++11` standard.

src/HipifyAction.h
```diff
@@ -42,6 +42,7 @@ class HipifyAction : public clang::ASTFrontendAction,
 private:
   ct::Replacements *replacements = nullptr;
   std::map<std::string, clang::SourceLocation> Ifndefs;
+  std::map<std::string, unsigned> InsertedVarCounter;
```
emankov: Where is this class member counter cleaned?

src/HipifyAction.cpp
```diff
@@ -2733,6 +2733,7 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
       std::string combinedArgText;
       bool hasInsertions = false;
       std::vector<hipify::CastInfo> insertionWarnings;
+      std::vector<std::string> insertionVarNames;
```
emankov: [Question] What about joining them in a single `std::pair` for synchronization reasons?

src/HipifyAction.cpp
ranapratap55: updated after adapting std::pair.

src/HipifyAction.cpp
ranapratap55: replaced with `std::pair` and restored range based `for` loop.

src/HipifyAction.h
ranapratap55: `InsertedVarCounter` resets implicitly per input file (via `ReplacementsFrontendActionFactory::create()`). 

Implemented `std::pair(FileID, baseName)` approach to give each source file in the translation unit its own independent counter.

emankov: > Implemented std::pair(FileID, baseName) approach to give each source file in the translation unit its own independent counter.

Is a scope actually a compilation unit (`FileID`)? What if there is more than one function in a single `FileID` to transform this way?

emankov: Tests should take scoping into account, too. It would need at least one more function besides `main`.

Additionally, a dummy header file (e.g., `#include "dummy_cudnn_call_header.h"`) might also be needed to prove that the `.h` and `.cu` file counters operate independently without polluting each other.

ranapratap55: Fixed it, thanks.

Traversed AST parent of the call to find out nearest `CompoundStmt`. Now, `main()` and `secondFn()` in the same file get independent counters. Added new test.

src/HipifyAction.h
```diff
@@ -42,6 +42,11 @@ class HipifyAction : public clang::ASTFrontendAction,
 private:
   ct::Replacements *replacements = nullptr;
   std::map<std::string, clang::SourceLocation> Ifndefs;
+  // Tracks how many times each inserted variable name (e.g., hipify_use_mask)
+  // has been used per C++ block scope, to avoid redefinition when the same
+  // API is called multiple times in the same scope. getRawEncoding() is
+  // globally unique per block across the entire translation unit.
+  std::map<std::pair<unsigned, std::string>, unsigned> InsertedVarCounter;
```
emankov: `FileID` is still needed to prevent loss of `translation-unit isolation`.
What about using `std::tuple` inside the map?

tests/unit_tests/synthetic/libraries/cudnn_insert_new_arg.cu
```diff
@@ -0,0 +1,64 @@
...
+  status = cudnnSetDropoutDescriptor(DropoutDescriptor, handle, dropout, states, reserveSpaceNumBytes, seed);
+
+  // Third call in same scope: _2 suffix.
+  // CHECK: bool hipify_use_mask_2 = {};
+  // CHECK-NEXT: bool hipify_state_evo_2 = {};
+  // CHECK-NEXT: miopenRNGType_t hipify_rng_mode_2 = {};
+  // CHECK-NEXT: status = miopenSetDropoutDescriptor(DropoutDescriptor, handle, dropout, states, reserveSpaceNumBytes, seed, hipify_use_mask_2, hipify_state_evo_2, hipify_rng_mode_2);
+  status = cudnnSetDropoutDescriptor(DropoutDescriptor, handle, dropout, states, reserveSpaceNumBytes, seed);
+
+  // CHECK: return 0;
```
emankov: Excessive check?

src/HipifyAction.cpp
emankov, line 2792: [Assumption] The `if-else` construct without curly braces in the source CUDA code will lead to syntactically incorrect HIP code, for instance:

```cpp
if (condition)
        status = cudnnSetDropoutDescriptor(desc, handle, dropout, states, size, seed);
    else
        status = CUDNN_STATUS_SUCCESS;
```
[ToDo] The `if-else` construct without curly braces should be presented in tessts.

src/HipifyAction.cpp
```diff
@@ -2742,14 +2742,26 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
         indent += *indentPtr++;
       for (auto c : cc.castMap) {
         if (c.second.castType == e_insert_new_argument) {
+          const std::string &baseName = c.second.constValToAddOrReplace;
+          unsigned scopeKey = 0;
+          auto parents = Result.Context->getParents(*call);
+          while (!parents.empty()) {
+            if (const auto *cs = parents[0].get<clang::CompoundStmt>()) {
+              scopeKey = cs->getLBracLoc().getRawEncoding();
+              break;
```
emankov: `AST` traverse here should cover all parents, not only the first one, which are very common in `template Instantiations`, `macros`, and `implicit conversions`.

src/HipifyAction.cpp
```diff
@@ -2742,14 +2742,26 @@ bool HipifyAction::cudaHostFuncCall(const mat::MatchFinder::MatchResult &Result)
         indent += *indentPtr++;
       for (auto c : cc.castMap) {
         if (c.second.castType == e_insert_new_argument) {
+          const std::string &baseName = c.second.constValToAddOrReplace;
+          unsigned scopeKey = 0;
+          auto parents = Result.Context->getParents(*call);
+          while (!parents.empty()) {
+            if (const auto *cs = parents[0].get<clang::CompoundStmt>()) {
```
emankov: [Performance] The expensive `AST` tree climb to find the `CompoundStmt` three separate times for the exact same function call.

emankov (changes requested):

src/HipifyAction.h
ranapratap55: Done. Changed the map type to` std::map<std::tuple<clang::FileID, unsigned, std::string>, unsigned>`.

tests/unit_tests/synthetic/libraries/cudnn_insert_new_arg.cu
ranapratap55: Fixed.

src/HipifyAction.cpp
ranapratap55: yes, thanks. This is a ToDo 1 from #2415, Added a `// TODO:` block in .cpp and test file.

src/HipifyAction.cpp
ranapratap55: Thanks for this. Fixed in the latest revision. 
Corrected to` parent.get<clang::CompoundStmt>()`, so every node at the current AST level is checked before ascending.

src/HipifyAction.cpp
ranapratap55: Thanks and fixed. 
Moved the logic outside the `for (auto c : cc.castMap)` loop and AST climb happens exactly once.

PR #2455, opened 2026-03-26
[HIPIFY][Perl][#1776][fix] Fix CUDA kernel launch legacy hipification

emankov:
**[Synopsis]**
+ Legacy hipification of CUDA kernel launch calls (under `--hip-kernel-execution-syntax`) is completely broken: `kernelName<<<gridDimensions, blockDimensions, sharedMemBytes, stream>>>(arguments)`

**[Solution]**
+ [Regex Pattern Flattening] Consolidated the CUDA launch parsing pattern into a single-line C++ string literal
+ [Memory-Safe Trimming] Introduced mutable variables in argument whitespace trimming to prevent silent Perl "Modification of a read-only value" runtime crashes on padded literals
+ [Robust AST Parsing] Implemented a character-by-character scanner to bypass commas nested within complex execution bounds correctly (e.g., `std::min(x, y)`)

**[Impact]**
+ Fully restores CUDA kernel launch syntax translation under the `--hip-kernel-execution-syntax` flag (legacy path).

**[Testing]**
+ `perl -c hipify-perl`
+ Test `complex_kernel_launches.cu`:
  - `hipify-clang` and `hipify-perl` give identical HIP output with and without `--hip-kernel-execution-syntax` option
+ [ToDo] Implement an automated `hipify-perl` test harness based on the existing `hipify-clang` one


ranapratap55 (approved):
LGTM.

PR #2467, opened 2026-03-30
[HIPIFY][fix] `LLVM upstream preparations` - Part 4 - `namespace` usage revision

ranapratap55:
Addresses `using namespace` violations in headers and main source(.cpp) files to match with LLVM coding standards. 

- Removed `using namespace llvm;`, `using namespace llvm;` and used `llvm::`, `std::` in header files.
- Added `using namespace llvm;` in main source(.cpp) files

### DIFF
--- a/src/CUDA2HIP_Doc.cpp
+++ b/src/CUDA2HIP_Doc.cpp
@@ -28,14 +28,22 @@ THE SOFTWARE.
 
 namespace doc {
 
-  using namespace std;
-  using namespace llvm;
+  using std::string;
+  using std::map;
+  using std::vector;
+  using std::unique_ptr;
+  using std::ostream;
+  using std::ofstream;
+  using std::stringstream;
+  using std::error_code;
+  using std::make_pair;
+  using std::endl;
 
-  typedef map<unsigned int, StringRef> sectionMap;
-  typedef map<StringRef, hipCounter> functionMap;
+  typedef map<unsigned int, llvm::StringRef> sectionMap;
+  typedef map<llvm::StringRef, hipCounter> functionMap;
   typedef functionMap typeMap;
-  typedef map<StringRef, cudaAPIversions> versionMap;
-  typedef map<StringRef, hipAPIversions> hipVersionMap;
+  typedef map<llvm::StringRef, cudaAPIversions> versionMap;
+  typedef map<llvm::StringRef, hipAPIversions> hipVersionMap;
   typedef map<llvm::StringRef, hipAPIChangedVersions> hipChangedVersionMap;
   typedef map<llvm::StringRef, cudaAPIChangedVersions> cudaChangedVersionMap;
   typedef map<llvm::StringRef, cudaAPIUnsupportedVersions> cudaUnsupportedVersionMap;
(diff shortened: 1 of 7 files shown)


src/CUDA2HIP_Doc.cpp
```diff
@@ -28,14 +28,22 @@ THE SOFTWARE.
 
 namespace doc {
 
-  using namespace std;
-  using namespace llvm;
```
emankov, line 32:
```cpp
using namespace llvm;
```
Inconsistent llvm namespace management, as the same directive was explicitly added to (or kept in) other source files.
The point is to remove `using namespace llvm` from header files. Utilizing `using namespace` within `.cpp` is safe. Whereas all `llvm::` prefixes are redundant.

src/CUDA2HIP_Doc.cpp
```diff
@@ -243,15 +250,15 @@ namespace doc {
...
+        EC = llvm::sys::fs::createTemporaryFile(file, getExtension(t), tmpFile);
         if (EC) {
-          errs() << "\n" << sHipify << sError << EC.message() << ": " << tmpFile << "\n";
+          llvm::errs() << "\n" << sHipify << sError << EC.message() << ": " << tmpFile << "\n";
           return false;
         }
         files.insert({ t, file });
         tmpFiles.insert({ t, tmpFile.str().str() });
-        streams.insert(make_pair(t, unique_ptr<ostream>(new ofstream(tmpFile.c_str(), ios_base::trunc))));
+        streams.insert(make_pair(t, unique_ptr<ostream>(new ofstream(tmpFile.c_str(), std::ios_base::trunc))));
```
emankov, line 261: Based on the change logic, `using std::ios_base;` should be utilized instead.

src/main.cpp
```diff
@@ -55,6 +55,8 @@ THE SOFTWARE.
 
 constexpr auto DEBUG_TYPE = "cuda2hip";
 
+using namespace llvm;
```
emankov, line 58: Based on the PR logic, all `llvm::` prefixes should be removed in this compilation unit as well.

src/CUDA2HIP_Doc.cpp
```diff
@@ -28,14 +28,22 @@ THE SOFTWARE.
...
+  using std::ostream;
+  using std::ofstream;
+  using std::stringstream;
+  using std::error_code;
+  using std::make_pair;
+  using std::endl;
 
-  typedef map<unsigned int, StringRef> sectionMap;
-  typedef map<StringRef, hipCounter> functionMap;
+  typedef map<unsigned int, llvm::StringRef> sectionMap;
```
emankov, line 42: Here and below, in the compilation unit, the `llvm::` prefixes are unnecessary (including the existing ones). Keeping `using namespace llvm;` is recommended instead.

emankov (changes requested):

src/CUDA2HIP_Doc.cpp
ranapratap55: Agreed and updated in the latest version.

src/CUDA2HIP_Doc.cpp
ranapratap55: Updated.

src/main.cpp
ranapratap55: Updated.

src/CUDA2HIP_Doc.cpp
ranapratap55: Updated.

kzhuravl (approved):

Issue #2584, opened 2026-06-12
[HIPIFY][CUB][test][failure] Failures on compiling `CUB` tests


emankov:
**[Failure]**
```cmd
1>  # | [HIPIFY] info: Actual CUDA SDK version is used by clang as   : 12.2
1>  # | [HIPIFY] info: Actual CUDA SDK version is treated by clang as: 12.2
1>  # |
1>  # | In file included from C:\Users\emankov\AppData\Local\Temp\lit-tmp-jf5jh0n9\cub_01.cu-505290.hip:8:
1>  # | In file included from C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.2.1/include\cub/cub.cuh:54:
1>  # | In file included from C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.2.1/include\cub\device/device_adjacent_difference.cuh:33:
1>  # | In file included from C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.2.1/include\cub/device/dispatch/dispatch_adjacent_difference.cuh:39:
1># | C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.2.1/include\thrust/system/cuda/detail/core/triple_chevron_launch.h(120,14): error G5FF4417D: kernel launch from __device__ or __global__ function requires relocatable device code (i.e. requires -fgpu-rdc)
1>  # |   120 |       return cudaLaunchDevice((void*)k,
1>  # |       |              ^
1>  # | 1 error generated when compiling for host.
1>  # | Error while processing C:\Users\emankov\AppData\Local\Temp\lit-tmp-jf5jh0n9\cub_01.cu-505290.hip.
```
**[Affected tests]**
+ cub_01.cu
+ cub_02.cu
+ cub_03.cu

**[Affected CUDA]**
`11.0.1` - `12.2.1`

emankov:
Fixed by #2585

PR #2585, opened 2026-06-12
[HIPIFY][#2584][fix] Fix failed `CUB` tests

emankov:
**[Root Cause]**
```cpp
kernel launch from __device__ or __global__ function requires relocatable device code (i.e. requires -fgpu-rdc)
```
**[Solution]**
+ Provide two versions of `CUB` tests: with and without clang option `-fgpu-rdc`
+ [Reason] The `-fgpu-rdc` option was officially introduced in clang `8.0.0`


ranapratap55 (approved):
LGTM.

PR #2639, opened 2026-06-26
[HIPIFY][fix][LCOMP-2369] Exclude standalone-CUB tests on CUDA >= 12.6 (fixes cstdint.h build failure)

ranapratap55:
These tests force a **standalone CUB** include path. From CCCL >= 12.6 (CUDA >= 12.6), `thrust/detail/cstdint.h` was removed, so standalone CUB include will breaks. This is a dependency-version mismatch between standalone CUB and latest CUDA versions not a clang parsing limitation. It reproduces on any CUDA >= 12.6 regardless of the LLVM/clang version.

Extend the existing CUB exclusion guard to also cover the newer `cub_*_LLVM_8.cu` variants when CUDA >= 12.6, alongside the pre-existing gates for CUDA < 9 and `CUDA_CUB_ROOT_DIR=OFF`.

Tested on **CUDA 12.8.93 / LLVM 20.1.8**:

## Follow-up (separate)
A longer-term fix is to correct include ordering (e.g. `-idirafter`) so the CCCL-bundled CUB takes precedence rather than excluding the tests. That changes header resolution globally, so it should be validated across the full CUDA x LLVM matrix and restore CUB coverage on CUDA >= 12.6 before landing. Will create a separate issue to track.

### DIFF
--- a/tests/lit.cfg
+++ b/tests/lit.cfg
@@ -43,17 +43,24 @@ else:
     if config.cudnn_version_major >= 9:
         config.excludes.append('cudnn2miopen_before_9000.cu')
 
+# cub_0N.cu run on LLVM < 8; cub_0N_LLVM_8.cu run on LLVM >= 8.
+cub_tests_pre_llvm8 = ['cub_01.cu', 'cub_02.cu', 'cub_03.cu']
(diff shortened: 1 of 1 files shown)


ronlieb (approved):
Please create 2nd ticket for Evgeniy to provide fix to allow tests to be run

PR #2654, opened 2026-07-07
[HIPIFY][tests][refactor] Simplify test harness

emankov:
+ Get rid of old head-on solutions with `CUDA` root folders' includes in `lit.cfg`
+ [Testing] `Linux` and `Windows`


ranapratap55 (approved):
LGTM.

PR #2675, opened 2026-07-22
[HIPIFY][fix] Quote find output safely in bin wrapper scripts

arvindcheru:
## Motivation

The wrapper scripts pass find output through unquoted backtick substitution, which can split filenames on spaces/newlines and treat dash-prefixed names as options when calling hipify-perl / hipify-clang. T

Review follow-up restores documented legacy behavior (default to current directory, per-extension find order) while keeping the NUL-safe filename handling.

## Technical Details

Changes:
Replace `findcode*.sh $SEARCH_DIR` with find … -print0 + mapfile -d '' + "${files[@]}"
Quote "$SEARCH_DIR" and pass hipify/clang args via bash arrays
Fix SCRIPT_NANE → SCRIPT_NAME so -filter=custom uses findcode_custom.sh

Review fixes (commit 7ea40e4):
Default to current directory when no path is given (set -- . in findcode*.sh; SEARCH_DIR=${SEARCH_DIR:-.} in wrappers; ${1:-.} in finduncodep.sh)
Restore per-extension find order using separate find … -print0 calls (same grouping as original scripts)

## Test Plan

- [x] `bash -n bin/*.sh` — shell syntax
- [x] `verify-wrapper-scripts.sh` — no-arg default, empty path, extension order, security filenames
- [x] **Functional** — `hipexamine-perl.sh tests/unit_tests/samples` — hipify stats on real samples
- [x] **Filter** — `findcode_custom.sh` / `-filter=custom` — only `.cu` (not `.cuh`) selected; `SCRIPT_NAME` typo fixed
- [x] **Path with spaces** — `findcode.sh` / `hipexamine-perl.sh` on directory name containing spaces
- [ ] PR CI — HIPIFY GitHub checks (markdownlint only; wrapper scripts not in CI)

## Test Result

bash verify-wrapper-scripts.sh

=== bash -n ===
PASS
=== no-arg findcode (defaults to .) ===
PASS count=4
=== empty path (defaults to .) ===
PASS
=== extension order ===
PASS a.cu b.cpp c.cuh x.h 
=== security: spaced/newline/dash names stay single entries ===
PASS
=== finduncodep no arg ===
PASS count=3
All checks passed.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/bin/findcode.sh
+++ b/bin/findcode.sh
@@ -1,10 +1,16 @@
 #!/bin/bash
 
-SEARCH_DIRS=$@
+# Outputs matched CUDA/HIP source and header files, NUL-delimited.
 
-find $SEARCH_DIRS -name '*.cu' -o -name '*.CU'
-find $SEARCH_DIRS -name '*.cpp' -o -name '*.cxx' -o -name '*.c' -o -name '*.cc'
-find $SEARCH_DIRS -name '*.CPP' -o -name '*.CXX' -o -name '*.C' -o -name '*.CC'
-find $SEARCH_DIRS -name '*.cuh' -o -name '*.CUH'
-find $SEARCH_DIRS -name '*.h' -o -name '*.hpp' -o -name '*.hh' -o -name '*.inc' -o -name '*.inl' -o -name '*.hxx' -o -name '*.hdl'
-find $SEARCH_DIRS -name '*.H' -o -name '*.HPP' -o -name '*.HH' -o -name '*.INC' -o -name '*.INL' -o -name '*.HXX' -o -name '*.HDL'
+if [ "$#" -eq 0 ]; then
+  exit 0
+fi
+
+find "$@" \( \
+  -name '*.cu' -o -name '*.CU' -o \
+  -name '*.cpp' -o -name '*.cxx' -o -name '*.c' -o -name '*.cc' -o \
+  -name '*.CPP' -o -name '*.CXX' -o -name '*.C' -o -name '*.CC' -o \
+  -name '*.cuh' -o -name '*.CUH' -o \
+  -name '*.h' -o -name '*.hpp' -o -name '*.hh' -o -name '*.inc' -o -name '*.inl' -o -name '*.hxx' -o -name '*.hdl' -o \
+  -name '*.H' -o -name '*.HPP' -o -name '*.HH' -o -name '*.INC' -o -name '*.INL' -o -name '*.HXX' -o -name '*.HDL' \
+\) -print0
--- a/bin/findcode.sh
+++ b/bin/findcode.sh
@@ -2,15 +2,13 @@
 
 # Outputs matched CUDA/HIP source and header files, NUL-delimited.
 
-if [ "$#" -eq 0 ]; then
-  exit 0
+if [ "$#" -eq 0 ] || { [ "$#" -eq 1 ] && [ -z "$1" ]; }; then
+  set -- .
 fi
 
-find "$@" \( \
-  -name '*.cu' -o -name '*.CU' -o \
-  -name '*.cpp' -o -name '*.cxx' -o -name '*.c' -o -name '*.cc' -o \
-  -name '*.CPP' -o -name '*.CXX' -o -name '*.C' -o -name '*.CC' -o \
-  -name '*.cuh' -o -name '*.CUH' -o \
-  -name '*.h' -o -name '*.hpp' -o -name '*.hh' -o -name '*.inc' -o -name '*.inl' -o -name '*.hxx' -o -name '*.hdl' -o \
-  -name '*.H' -o -name '*.HPP' -o -name '*.HH' -o -name '*.INC' -o -name '*.INL' -o -name '*.HXX' -o -name '*.HDL' \
-\) -print0
+find "$@" \( -name '*.cu' -o -name '*.CU' \) -print0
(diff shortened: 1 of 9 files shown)


bin/findcode.sh
```diff
@@ -1,10 +1,16 @@
...
 
-find $SEARCH_DIRS -name '*.cu' -o -name '*.CU'
-find $SEARCH_DIRS -name '*.cpp' -o -name '*.cxx' -o -name '*.c' -o -name '*.cc'
-find $SEARCH_DIRS -name '*.CPP' -o -name '*.CXX' -o -name '*.C' -o -name '*.CC'
-find $SEARCH_DIRS -name '*.cuh' -o -name '*.CUH'
-find $SEARCH_DIRS -name '*.h' -o -name '*.hpp' -o -name '*.hh' -o -name '*.inc' -o -name '*.inl' -o -name '*.hxx' -o -name '*.hdl'
-find $SEARCH_DIRS -name '*.H' -o -name '*.HPP' -o -name '*.HH' -o -name '*.INC' -o -name '*.INL' -o -name '*.HXX' -o -name '*.HDL'
+if [ "$#" -eq 0 ]; then
+  exit 0
+fi
```
emankov: Here and below. It breaks the default logic without arguments.

bin/findcode.sh
```diff
@@ -1,10 +1,16 @@
...
+
+find "$@" \( \
+  -name '*.cu' -o -name '*.CU' -o \
+  -name '*.cpp' -o -name '*.cxx' -o -name '*.c' -o -name '*.cc' -o \
+  -name '*.CPP' -o -name '*.CXX' -o -name '*.C' -o -name '*.CC' -o \
+  -name '*.cuh' -o -name '*.CUH' -o \
+  -name '*.h' -o -name '*.hpp' -o -name '*.hh' -o -name '*.inc' -o -name '*.inl' -o -name '*.hxx' -o -name '*.hdl' -o \
+  -name '*.H' -o -name '*.HPP' -o -name '*.HH' -o -name '*.INC' -o -name '*.INL' -o -name '*.HXX' -o -name '*.HDL' \
+\) -print0
```
emankov: Here and below. The sorting logic is changed from `per file extension` to `per parent directory`

bin/hipconvertinplace-perl.sh
```diff
@@ -27,9 +27,14 @@ SCRIPT_NAME=findcode_sources.sh
...
 elif [ "$2" = "-filter=custom" ]
 then
-SCRIPT_NANE=findcode_custom.sh
+SCRIPT_NAME=findcode_custom.sh
 shift
 fi
 shift
 
-$SCRIPT_DIR/hipify-perl -inplace -print-stats "$@" `$SCRIPT_DIR/$SCRIPT_NAME $SEARCH_DIR`
+mapfile -d '' -t files < <("$SCRIPT_DIR/$SCRIPT_NAME" "$SEARCH_DIR")
```
emankov, line 40: Here and below. Previously, if no arguments are provided, the script searches the current directory. With the change, `find ""` leads to error.

emankov (changes requested):

bin/findcode.sh
arvindcheru: Thanks — agreed. exit 0 on empty args was wrong; the legacy scripts search the current directory when no path is given.

Fixed in 7ea40e4: replaced early exit with set -- . when there are no args (or a single empty arg), so find "$@" … -print0 searches . as before. Applied to findcode.sh, findcode_headers.sh, findcode_sources.sh, and findcode_custom.sh.

bin/findcode.sh
arvindcheru: Thank you for comment - Merging into one find changed output from extension-group order to directory-walk order, which is not what the legacy scripts did.

Fixed in 7ea40e4: restored separate find calls per extension group (same grouping as the original scripts), each with -print0. Wrappers still use mapfile -d '', so filename handling stays safe while output order matches legacy behavior.

bin/hipconvertinplace-perl.sh
arvindcheru: Thank you for the comments, 

Fixed in 7ea40e4: default empty search path to . in all wrapper scripts (SEARCH_DIR=${SEARCH_DIR:-.} in hipconvertinplace-perl.sh, hipconvertinplace.sh, hipexamine-perl.sh, hipexamine.sh) and ${1:-.} in finduncodep.sh. findcode*.sh also treats a single empty argument as . via set -- ..

bin/hipconvertinplace.sh
```diff
@@ -8,18 +8,25 @@
 
 SCRIPT_DIR="$(dirname "$(realpath "$0")")"
 BIN_DIR="$SCRIPT_DIR/../../bin"
-SEARCH_DIR=$1
+SEARCH_DIR=${1:-.}
+shift
```
emankov: `shift count out of range`, if no arguments.
Could you please check the `shift` in the rest of the scripts, too?

emankov (changes requested):

bin/hipconvertinplace.sh
arvindcheru: Thank you for the comment, fix updated in latest version

emankov (approved):
LGTM, thank you!

PR #2682, opened 2026-07-28
[HIPIFY][CUB][tests][fix] Exclude `CUB` tests for `CUDA 13`, too

emankov:
+ [Reason] `<thrust/detail/cstdint.h>`, removed from `CCCL` in `CUDA 12.6`


ranapratap55 (approved):
LGTM.

PR #2695, opened 2026-07-30
[HIPIFY][build][CMake][refactor] Transition to Modern CMake target properties

emankov:
+ Removed global `include_directories` and `add_definitions` directives to prevent implicit scope pollution.
+ Implemented `target_include_directories` with `PRIVATE` scope to strictly isolate LLVM and Clang header search paths.
+ Added safe parsing of the `LLVM_DEFINITIONS` string via `separate_arguments`, applying them through `target_compile_options`.
+ Ensured strict build-level dependency encapsulation for the `hipify-clang` binary target.

**[Problems Solved]**
+ Scope Pollution: Legacy directory-level configuration commands forced LLVM dependencies to propagate globally. This created architectural risks of build conflicts when integrating the `hipify-clang` target as part of other SW.
+ Compiler Syntax Errors: Directly passing unparsed, space-delimited LLVM macro strings into modern target properties caused build failures due to macro flag prefix duplication.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,10 +57,6 @@ if (NOT HIPIFY_CLANG_TESTS_ONLY)
     return()
   endif()
 
-  include_directories(${CLANG_INCLUDE_DIRS})
-  include_directories(${LLVM_INCLUDE_DIRS})
-  add_definitions(${LLVM_DEFINITIONS})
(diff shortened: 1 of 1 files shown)


ranapratap55 (approved):
LGTM.

PR #2705, opened 2026-08-12
[HIPIFY] Add LLVM lib dir to hipify-clang RPATH

### LINKED ISSUE #2704: [HIPIFY] hipify-clang fails to start: RUNPATH misses LLVM lib dir in versioned ROCm layout (libLLVM.so.23.0git not found)
## Problem

`hipify-clang` from the ROCm 7.14 deb packages cannot start, because its `RUNPATH` does not contain the directory that ships `libLLVM.so.23.0git`:

```
$ hipify-clang --version
hipify-clang: error while loading shared libraries: libLLVM.so.23.0git: cannot open shared object file: No such file or directory
$ echo $?
127

$ ldd $(readlink -f $(command -v hipify-clang)) | grep -i llvm
	libLLVM.so.23.0git => not found
```

The binary's `RUNPATH` is a single relative entry:

```
$ readelf -d /opt/rocm/core-7.14/bin/hipify-clang | grep RUNPATH
 0x000000000000001d (RUNPATH)  Library runpath: [$ORIGIN/../lib]
```

With `$ORIGIN` = `/opt/rocm/core-7.14/bin`, that resolves to `/opt/rocm/core-7.14/lib`, but the library is one level deeper:

```
$ ls /opt/rocm/core-7.14/lib/libLLVM.so.23.0git
ls: cannot access ...: No such file or directory

$ find /opt/rocm -name 'libLLVM.so.23.0git'
/opt/rocm/core-7.14/lib/llvm/lib/libLLVM.so.23.0git
```

Confirming that the directory is the only thing missing:

```
$ LD_LIBRARY_PATH=/opt/rocm/core-7.14/lib/llvm/lib hipify-clang --version
AMD LLVM version 23.0.0git
  Optimized build.
```

`/opt/rocm/lib` is also absent from the loader cache on this system (see ROCm/TheRock#6716), so there is no fallback and the failure is unconditional on a stock install.

## Root cause

`CMakeLists.txt` hardcodes the RPATH one directory short of the LLVM library dir:

https://github.com/ROCm/HIPIFY/blob/amd-develop/CMakeLists.txt#L250-L254

```cmake
  if(UNIX)
    # Get rid of any RPATH definations already.
    set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "")
    # Set RPATH for the binary.
    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,\\$ORIGIN/../lib" )
```

`$ORIGIN/../lib` is correct for a flat layout where `libLLVM` sits directly in `<prefix>/lib`. It is wrong for the versioned packaging layout now used by ROCm, where the binary is at `core-X.Y/bin` and LLVM is installed under `core-X.Y/lib/llvm/lib`. Note also that `INSTALL_RPATH ""` on the preceding line discards whatever CMake would otherwise have derived from `target_link_directories(... ${LLVM_LIBRARY_DIRS})` at line 97, so the hardcoded value is the only thing left.

A fix would be to include the actual LLVM library directory, e.g. adding `$ORIGIN/../lib/llvm/lib` to the RPATH list, or deriving it from `${LLVM_LIBRARY_DIRS}` relative to the install prefix rather than hardcoding a single relative path.

## Environment

- Ubuntu 26.04, kernel 7.0
- ROCm 7.14.0~pre3 from `repo.radeon.com/rocmradeon/apt/26.13`, package `amdrocm-hipify7.14`
- gfx1100 (RX 7900 XTX)

## Note

This looks like a recurrence of the same class of bug as #678 (`Wrong RUNPATH in hipify-clang executable`) and #1959 (RPATH workaround on RHEL10), both previously fixed. `hipify-perl` is unaffected — it is a Perl script with no shared-library dependency — so only the clang-based tool is broken.

Samiii777:
Fixes #2704.

## Problem

`hipify-clang` from the ROCm 7.14 packages cannot start, because its `RUNPATH` does not include the directory that actually ships `libLLVM`:

```
$ hipify-clang --version
hipify-clang: error while loading shared libraries: libLLVM.so.23.0git: cannot open shared object file: No such file or directory

$ readelf -d /opt/rocm/core-7.14/bin/hipify-clang | grep RUNPATH
 0x000000000000001d (RUNPATH)  Library runpath: [$ORIGIN/../lib]

$ find /opt/rocm -name 'libLLVM.so.23.0git'
/opt/rocm/core-7.14/lib/llvm/lib/libLLVM.so.23.0git
```

`$ORIGIN/../lib` resolves to `<prefix>/lib`, but in the versioned ROCm layout LLVM is installed one level deeper under `<prefix>/lib/llvm/lib`. `/opt/rocm/lib` is also not registered with the dynamic loader (ROCm/TheRock#6716), so there is no fallback and the failure is unconditional on a stock install.

## Change

Append `$ORIGIN/../lib/llvm/lib` to the RPATH. The original entry is kept first, so a flat layout that installs `libLLVM` directly into `<prefix>/lib` continues to resolve exactly as before.

## Test

Building HIPIFY needs a full LLVM/Clang dev tree, so I verified the RPATH mechanism directly by reproducing the install layout (`bin/`, `lib/`, `lib/llvm/lib/`) and linking a binary against a stand-in library placed where `libLLVM` actually ships:

| case | RPATH | result |
|---|---|---|
| current upstream | `$ORIGIN/../lib` | fails — reproduces the reported error |
| this PR | `$ORIGIN/../lib:$ORIGIN/../lib/llvm/lib` | resolves with no `LD_LIBRARY_PATH` |
| flat layout regression guard | `$ORIGIN/../lib:$ORIGIN/../lib/llvm/lib`, lib in `../lib` | still resolves |

3 passed, 0 failed. The first row is a control: it must keep failing, otherwise the test would be vacuous.

Independently, adding only that directory to the search path is sufficient to make the shipped binary work:

```
$ LD_LIBRARY_PATH=/opt/rocm/core-7.14/lib/llvm/lib hipify-clang --version
AMD LLVM version 23.0.0git
  Optimized build.
```

## Environment

Ubuntu 26.04, ROCm 7.14.0~pre3 (`amdrocm-hipify7.14`) from `repo.radeon.com/rocmradeon/apt/26.13`, gfx1100.

## Note

Opened as a draft. `hipify-perl` is unaffected (Perl script, no shared-library dependency). Related earlier fixes in this area: #678 and #1959.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,20 +256,37 @@ if (NOT HIPIFY_CLANG_TESTS_ONLY)
     # hipify-clang must therefore be able to find libLLVM at runtime, wherever
     # the consumed LLVM package installed it, hence deriving the RPATH below
     # from LLVM_LIBRARY_DIRS instead of assuming a fixed layout.
-    set(HIPIFY_RPATH_DIRS "\\$ORIGIN/../lib" "\\$ORIGIN/../lib/llvm/lib")
-    foreach(_llvm_lib_dir ${LLVM_LIBRARY_DIRS})
+    #
+    # $ORIGIN/../lib covers the flat layout; the versioned ROCm layout installs
+    # the binary in core-X.Y/bin and libLLVM in core-X.Y/lib/llvm/lib.
+    set(HIPIFY_INSTALL_RPATH "\$ORIGIN/../lib" "\$ORIGIN/../lib/llvm/lib")
+    foreach(_llvm_lib_dir IN LISTS LLVM_LIBRARY_DIRS)
       file(RELATIVE_PATH _llvm_lib_rel "${CMAKE_INSTALL_PREFIX}" "${_llvm_lib_dir}")
-      if(_llvm_lib_rel AND NOT IS_ABSOLUTE "${_llvm_lib_rel}" AND NOT _llvm_lib_rel MATCHES "^\\.\\.")
-        list(APPEND HIPIFY_RPATH_DIRS "\\$ORIGIN/../${_llvm_lib_rel}")
+      if(NOT _llvm_lib_rel OR IS_ABSOLUTE "${_llvm_lib_rel}" OR _llvm_lib_rel MATCHES "^\\.\\.")
+        # Outside CMAKE_INSTALL_PREFIX, so deliberately left out of the install
+        # RPATH: an absolute build-machine path must not be baked into a
+        # redistributable binary. The build-tree binary still resolves libLLVM
+        # through the BUILD_RPATH that llvm_setup_rpath() derives from
+        # LLVM_LIBRARY_DIR for exactly this out-of-prefix case.
+        message(STATUS "   - LLVM lib dir outside install prefix, omitted from hipify-clang install RPATH: ${_llvm_lib_dir}")
+      elseif(_llvm_lib_rel MATCHES ":")
+        # ':' separates entries in an ELF RPATH and cannot be escaped.
+        message(WARNING "LLVM lib dir contains ':' and cannot be represented in an RPATH, omitted from hipify-clang install RPATH: ${_llvm_lib_dir}")
+      else()
+        list(APPEND HIPIFY_INSTALL_RPATH "\$ORIGIN/../${_llvm_lib_rel}")
       endif()
     endforeach()
-    list(REMOVE_DUPLICATES HIPIFY_RPATH_DIRS)
-    list(JOIN HIPIFY_RPATH_DIRS ":" HIPIFY_RPATH)
+    list(REMOVE_DUPLICATES HIPIFY_INSTALL_RPATH)
 
-    # Get rid of any RPATH definations already.
-    set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "")
-    # Set RPATH for the binary.
-    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,${HIPIFY_RPATH}" )
+    # Set the RPATH via INSTALL_RPATH rather than a hand-built LINK_FLAGS
+    # string. LINK_FLAGS is a single unquoted string, so any path containing a
+    # space would split the -Wl,--rpath argument, and assigning it here would
+    # also discard the -Wl,-rpath-link (GNU ld) and -Wl,-z,origin (FreeBSD)
+    # flags that llvm_setup_rpath() appended from add_llvm_executable().
+    # INSTALL_RPATH is a real list that CMake quotes and joins itself, and
+    # setting it replaces the value llvm_setup_rpath() had installed.
+    set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "${HIPIFY_INSTALL_RPATH}")
+    target_link_options(hipify-clang PRIVATE "LINKER:--enable-new-dtags")
 
     set(CPACK_PACKAGING_INSTALL_PREFIX "/opt/rocm" CACHE PATH "HIP Package Installation Path")
     set(BUILD_DIR ${CMAKE_CURRENT_BINARY_DIR}/packages/hipify-clang)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,7 +251,7 @@ if (NOT HIPIFY_CLANG_TESTS_ONLY)
     # Get rid of any RPATH definations already.
     set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "")
     # Set RPATH for the binary.
-    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,\\$ORIGIN/../lib" )
+    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,\\$ORIGIN/../lib:\\$ORIGIN/../lib/llvm/lib" )
 
     set(CPACK_PACKAGING_INSTALL_PREFIX "/opt/rocm" CACHE PATH "HIP Package Installation Path")
     set(BUILD_DIR ${CMAKE_CURRENT_BINARY_DIR}/packages/hipify-clang)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,10 +248,28 @@ if (NOT HIPIFY_CLANG_TESTS_ONLY)
   endif()
 
   if(UNIX)
+    # hipify-clang cannot be statically linked against LLVM/clang on platforms
+    # (e.g. TheRock's Linux build) where LLVM_LINK_LLVM_DYLIB=ON: clang's static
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -251,7 +251,7 @@ if (NOT HIPIFY_CLANG_TESTS_ONLY)
     # Get rid of any RPATH definations already.
     set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "")
     # Set RPATH for the binary.
-    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,\\$ORIGIN/../lib" )
+    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,\\$ORIGIN/../lib:\\$ORIGIN/../lib/llvm/lib" )
```
emankov: `hipify-clang` should be statically linked against `LLVM` and `clang`. Dynamic linking is undocumented and unsupported behavior.

emankov (changes requested):

CMakeLists.txt
```diff
@@ -248,10 +248,28 @@ if (NOT HIPIFY_CLANG_TESTS_ONLY)
...
+    # libraries carry a transitive dependency on libLLVM in that configuration,
+    # so mixing them with statically-linked LLVM components causes ODR
+    # violations ("duplicate cl::opt registered") at runtime - see #1840/#1873.
+    # hipify-clang must therefore be able to find libLLVM at runtime, wherever
+    # the consumed LLVM package installed it, hence deriving the RPATH below
+    # from LLVM_LIBRARY_DIRS instead of assuming a fixed layout.
+    set(HIPIFY_RPATH_DIRS "\\$ORIGIN/../lib" "\\$ORIGIN/../lib/llvm/lib")
+    foreach(_llvm_lib_dir ${LLVM_LIBRARY_DIRS})
+      file(RELATIVE_PATH _llvm_lib_rel "${CMAKE_INSTALL_PREFIX}" "${_llvm_lib_dir}")
+      if(_llvm_lib_rel AND NOT IS_ABSOLUTE "${_llvm_lib_rel}" AND NOT _llvm_lib_rel MATCHES "^\\.\\.")
```
ranapratap55: This guard only passes when LLVM's libdir sits inside `CMAKE_INSTALL_PREFIX`.

CMakeLists.txt
```diff
@@ -248,10 +248,28 @@ if (NOT HIPIFY_CLANG_TESTS_ONLY)
...
+      endif()
+    endforeach()
+    list(REMOVE_DUPLICATES HIPIFY_RPATH_DIRS)
+    list(JOIN HIPIFY_RPATH_DIRS ":" HIPIFY_RPATH)
+
     # Get rid of any RPATH definations already.
     set_target_properties(hipify-clang PROPERTIES INSTALL_RPATH "")
     # Set RPATH for the binary.
-    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,\\$ORIGIN/../lib" )
+    set_target_properties(hipify-clang PROPERTIES LINK_FLAGS "-Wl,--enable-new-dtags -Wl,--rpath,${HIPIFY_RPATH}" )
```
ranapratap55: `LINK_FLAGS` is a single string and `set_target_properties` replaces it. So a space in any path splits the `-Wl,--rpath` argument and a `:` can corrupt the list.

ranapratap55 (changes requested):

CMakeLists.txt
Samiii777: Fixed in 0334277. Reproduced the split first — with a space in a path the generated link line was `-Wl,--rpath,\$ORIGIN/../lib:\$ORIGIN/../my lib/llvm/lib`, unquoted, so `my lib/llvm/lib` became a separate linker argument.

One further consequence worth noting: `add_llvm_executable()` calls `llvm_setup_rpath()`, which *appends* `-Wl,-rpath-link,<LLVM_LIBRARY_OUTPUT_INTDIR>` (GNU ld) and `-Wl,-z,origin` (FreeBSD) to `LINK_FLAGS`. Assigning `LINK_FLAGS` here discarded those — pre-existing, not introduced by this PR.

It now uses the `INSTALL_RPATH` property (a real CMake list that CMake quotes and joins itself) plus `target_link_options(hipify-clang PRIVATE "LINKER:--enable-new-dtags")`, which is additive rather than clobbering. Setting `INSTALL_RPATH` also supersedes the value `llvm_setup_rpath()` installed, so the preceding `INSTALL_RPATH ""` line is no longer needed and is gone. A libdir containing `:` is skipped with a warning, since `:` separates entries in an ELF RPATH and cannot be escaped.

Verified the generated entries for in-prefix, out-of-prefix, space- and colon-containing library dirs; the space case now emits a correctly quoted `-add_rpath "$ORIGIN/../llvm 20/lib"` where the previous `LINK_FLAGS` form split the argument.

CMakeLists.txt
Samiii777: Correct, and that is intentional: a libdir outside `CMAKE_INSTALL_PREFIX` is an absolute build-machine path, which should not be baked into a redistributable binary.

That case is already covered elsewhere. `llvm_setup_rpath()` appends `LLVM_LIBRARY_DIR` to **BUILD_RPATH** precisely when `LLVM_INSTALL_PREFIX` differs from `CMAKE_INSTALL_PREFIX`, so the build-tree binary — the one developers and `test-hipify-clang` actually run — still resolves `libLLVM`. Only the installed binary is restricted to relative `$ORIGIN` entries.

It no longer fails silently, though: out-of-prefix dirs are now reported at STATUS level rather than being dropped without a trace. If you would rather the installed binary also carry the absolute path for non-packaged installs, I am happy to add that behind an option.

CMakeLists.txt
Samiii777: Static linking is not available in the configuration where this bug occurs.

TheRock sets `LLVM_BUILD_LLVM_DYLIB ON` / `LLVM_LINK_LLVM_DYLIB ON` unconditionally for Linux in [`compiler/pre_hook_amd-llvm.cmake`](https://github.com/ROCm/TheRock/blob/main/compiler/pre_hook_amd-llvm.cmake), which is why the shipped ROCm 7.14 package contains `libLLVM.so.23.0git` (#2704). In that mode clang's static libraries carry a transitive dependency on `libLLVM`, so combining them with statically linked LLVM components reintroduces the ODR "duplicate cl::opt registered" failures that #1840 / #1873 fixed — which is exactly why `hipify-clang` links its LLVM components via `LLVM_LINK_COMPONENTS` today.

So making `hipify-clang` statically linked would require changing TheRock's LLVM build configuration first, which is outside this repository. This PR does not add or extend dynamic linking; it only corrects where the binary looks for the `libLLVM` it is already linked against, which is why it is scoped to the RPATH.

If you do want to pursue static linking against `LLVM`/`clang`, that looks like a separate piece of work spanning HIPIFY and TheRock — should I open an issue for it?

